PR #60, opened 2021-09-03
feat(common): APPEX-167 Fix MySQL support for generic mysql servers

bc-zachary:
## What?
Adds a check for MySQL env vars before falling back to ClearDB for connection. 

## Why?
Jira: [APPEX-167](https://jira.bigcommerce.com/browse/APPEX-167)

...

## Testing / Proof
Verified locally
...

### DIFF
--- a/lib/dbs/mysql.ts
+++ b/lib/dbs/mysql.ts
@@ -2,9 +2,17 @@ import * as mysql from 'mysql';
 import { promisify } from 'util';
 import { SessionProps, StoreData } from '../../types';
 
+const MYSQL_CONFIG = {
+    host: process.env.MYSQL_HOST,
+    database: process.env.MYSQL_DATABASE,
+    user: process.env.MYSQL_USERNAME,
+    password: process.env.MYSQL_PASSWORD,
+    ...(process.env.MYSQL_PORT && { port: process.env.MYSQL_PORT }),
+};
+
 // For use with Heroku ClearDB
 // Other mysql: https://www.npmjs.com/package/mysql#establishing-connections
-const connection = mysql.createConnection(process.env.CLEARDB_DATABASE_URL);
+const connection = mysql.createConnection(process.env.CLEARDB_DATABASE_URL ? process.env.CLEARDB_DATABASE_URL : MYSQL_CONFIG);
 const query = promisify(connection.query.bind(connection));
 
 // Use setUser for storing global user data (persists between installs)
--- a/scripts/db.js
+++ b/scripts/db.js
@@ -1,7 +1,15 @@
 const mysql = require('mysql');
 const util = require('util');
 
-const connection = mysql.createConnection(process.env.CLEARDB_DATABASE_URL);
+const MYSQL_CONFIG = {
+    host: process.env.MYSQL_HOST,
+    database: process.env.MYSQL_DATABASE,
+    user: process.env.MYSQL_USERNAME,
+    password: process.env.MYSQL_PASSWORD,
+    ...(process.env.MYSQL_PORT && { port: process.env.MYSQL_PORT }),
+};
+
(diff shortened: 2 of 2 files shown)


tylerbroadway (approved):

lib/constants.ts
```diff
@@ -0,0 +1,7 @@
+export const MYSQL_CONFIG = {
```
bc-chaz, line 1: I don't think we need a constants file for the env variables - we should just be able to reference the `env`... by doing this we'll have to maintain the variables in two places: here and inside of .env

lib/dbs/mysql.ts
```diff
@@ -1,10 +1,11 @@
 import * as mysql from 'mysql';
 import { promisify } from 'util';
+import { MYSQL_CONFIG } from '@lib/constants';
 import { SessionProps, StoreData } from '../../types';
 
 // For use with Heroku ClearDB
 // Other mysql: https://www.npmjs.com/package/mysql#establishing-connections
-const connection = mysql.createConnection(process.env.CLEARDB_DATABASE_URL);
+const connection = mysql.createConnection(MYSQL_CONFIG.host ? MYSQL_CONFIG : process.env.CLEARDB_DATABASE_URL);
```
bc-chaz, line 8: it would probably be easier to follow to either build a url here or build the config object in this file

bc-chaz: we might want to instead check if `CLEARDB_DATABASE_URL` is defined first - since we have a deploy to Heroku button, we might want to give preference to clearDB

scripts/db.js
```diff
@@ -1,7 +1,8 @@
 const mysql = require('mysql');
 const util = require('util');
+import { MYSQL_CONFIG } from '@lib/constants';
```
bc-chaz, line 3: I don't believe this file exists anymore... I'd just port over the object here

scripts/db.js
```diff
@@ -1,7 +1,8 @@
...
 
-const connection = mysql.createConnection(process.env.CLEARDB_DATABASE_URL);
+const connection = mysql.createConnection(MYSQL_CONFIG.host ? MYSQL_CONFIG : process.env.CLEARDB_DATABASE_URL);
```
bc-chaz, line 5: same here - let's check if `CLEARDB_DATABASE_URL` since that should take precedence

lib/dbs/mysql.ts
```diff
@@ -2,9 +2,17 @@ import * as mysql from 'mysql';
 import { promisify } from 'util';
 import { SessionProps, StoreData } from '../../types';
 
+const MYSQL_CONFIG = {
+  host: process.env.MYSQL_HOST,
+  database: process.env.MYSQL_DATABASE,
+  user: process.env.MYSQL_USERNAME,
+  password: process.env.MYSQL_PASSWORD,
+  port: process.env.MYSQL_PORT,
```
bc-chaz, line 10: ahhh - I just noticed that we listed the port as optional in the `.env-sample` - https://github.com/bigcommerce/sample-app-nodejs/blob/main/.env-sample#L31

we might have to optionally add this property to the object:
```jsx
{
    ...(port && { port: MYSQL_PORT }),
}
```

bc-chaz: also, I believe we're using `4 space` indenting in this repo

scripts/db.js
```diff
@@ -1,7 +1,15 @@
 const mysql = require('mysql');
 const util = require('util');
 
-const connection = mysql.createConnection(process.env.CLEARDB_DATABASE_URL);
+const MYSQL_CONFIG = {
+  host: process.env.MYSQL_HOST,
+  database: process.env.MYSQL_DATABASE,
+  user: process.env.MYSQL_USERNAME,
+  password: process.env.MYSQL_PASSWORD,
+  port: process.env.MYSQL_PORT,
```
bc-chaz, line 9: let's update this to be optional as well

bc-chaz (approved):

PR #61, opened 2021-09-15
feat(common): APPEX-78 Add product list pagination to sample app

bc-zachary:
## What?
Adds pagination to the product list in sample app including selectable items per page.

Screen Shot 2021-09-15 at 4 06 57 PM

## Why?
Jira: [APPEX-78](https://jira.bigcommerce.com/browse/APPEX-78)
...

## Testing / Proof
Tests passing and verified locally. 
...

### DIFF
--- a/lib/hooks.ts
+++ b/lib/hooks.ts
@@ -1,9 +1,9 @@
 import useSWR from 'swr';
 import { useSession } from '../context/session';
-import { ErrorProps, ListItem } from '../types';
+import { ErrorProps, ListItem, QueryParams } from '../types';
 
-async function fetcher(url: string, encodedContext: string) {
-    const res = await fetch(`${url}?context=${encodedContext}`);
+async function fetcher(url: string, query: string) {
+    const res = await fetch(`${url}?${query}`);
 
     // If the status code is not in the range 200-299, throw an error
     if (!res.ok) {
@@ -19,9 +19,10 @@ async function fetcher(url: string, encodedContext: string) {
 // Reusable SWR hooks
 // https://swr.vercel.app/
 export function useProducts() {
-    const encodedContext = useSession()?.context;
+    const { context } = useSession();
+    const params = new URLSearchParams({ context }).toString();
     // Request is deduped and cached; Can be shared across components
-    const { data, error } = useSWR(encodedContext ? ['/api/products', encodedContext] : null, fetcher);
+    const { data, error } = useSWR(context ? ['/api/products', params] : null, fetcher);
 
     return {
         summary: data,
@@ -30,13 +31,16 @@ export function useProducts() {
     };
 }
 
-export function useProductList() {
-    const encodedContext = useSession()?.context;
+export function useProductList(query?: QueryParams) {
+    const { context } = useSession();
+    const params = new URLSearchParams({ ...query, context }).toString();
+
     // Use an array to send multiple arguments to fetcher
-    const { data, error, mutate: mutateList } = useSWR(encodedContext ? ['/api/products/list', encodedContext] : null, fetcher);
+    const { data, error, mutate: mutateList } = useSWR(context ? ['/api/products/list', params] : null, fetcher);
 
     return {
-        list: data,
+        list: data?.data,
+        meta: data?.meta,
         isLoading: !data && !error,
         error,
         mutateList,
@@ -44,10 +48,11 @@ export function useProductList() {
 }
 
 export function useProductInfo(pid: number, list: ListItem[]) {
-    const encodedContext = useSession()?.context;
+    const { context } = useSession();
+    const params = new URLSearchParams({ context }).toString();
     const product = list.find(item => item.id === pid);
     // Conditionally fetch product if it doesn't exist in the list (e.g. deep linking)
-    const { data, error } = useSWR(!product && encodedContext ? [`/api/products/${pid}`, encodedContext] : null, fetcher);
+    const { data, error } = useSWR(!product && context ? [`/api/products/${pid}`, params] : null, fetcher);
 
     return {
         product: product ?? data,
--- a/pages/api/products/list.ts
+++ b/pages/api/products/list.ts
@@ -1,17 +1,16 @@
 import { NextApiRequest, NextApiResponse } from 'next';
+import { URLSearchParams } from 'url';
 import { bigcommerceClient, getSession } from '../../../lib/auth';
 
 export default async function list(req: NextApiRequest, res: NextApiResponse) {
     try {
         const { accessToken, storeHash } = await getSession(req);
         const bigcommerce = bigcommerceClient(accessToken, storeHash);
-        // Optional: pass in API params here
-        const params = [
-            'limit=11',
-        ].join('&');
+        const { page, limit, sort, direction } = req.query;
+        const params = new URLSearchParams({ page, limit, ...(sort && {sort, direction}) }).toString();
 
-        const { data } = await bigcommerce.get(`/catalog/products?${params}`);
-        res.status(200).json(data);
+        const response = await bigcommerce.get(`/catalog/products?${params}`);
+        res.status(200).json(response);
     } catch (error) {
         const { message, response } = error;
         res.status(response?.status || 500).json({ message });
--- a/pages/products/index.tsx
+++ b/pages/products/index.tsx
@@ -1,22 +1,42 @@
-import { Button, Dropdown, Panel, Small, StatefulTable, Link as StyledLink } from '@bigcommerce/big-design';
+import { Button, Dropdown, Panel, Small, Link as StyledLink, Table, TableSortDirection } from '@bigcommerce/big-design';
 import { MoreHorizIcon } from '@bigcommerce/big-design-icons';
 import Link from 'next/link';
 import { useRouter } from 'next/router';
-import { ReactElement } from 'react';
+import { ReactElement, useState } from 'react';
 import ErrorMessage from '../../components/error';
 import Loading from '../../components/loading';
 import { useProductList } from '../../lib/hooks';
 import { TableItem } from '../../types';
 
 const Products = () => {
+    const [itemsPerPage, setItemsPerPage] = useState(10);
+    const [currentPage, setCurrentPage] = useState(1);
+    const [columnHash, setColumnHash] = useState('');
+    const [direction, setDirection] = useState<TableSortDirection>('ASC');
     const router = useRouter();
-    const { error, isLoading, list = [] } = useProductList();
+    const { error, isLoading, list = [], meta = {} } = useProductList({
+      page: String(currentPage),
+      limit: String(itemsPerPage),
+      ...(columnHash && { sort: columnHash }),
+      ...(columnHash && { direction: direction.toLowerCase() }),
+    });
+    const itemsPerPageOptions = [10, 20, 50, 100];
     const tableItems: TableItem[] = list.map(({ id, inventory_level: stock, name, price }) => ({
         id,
         name,
         price,
         stock,
     }));
+
+    const onItemsPerPageChange = newRange => {
+        setCurrentPage(1);
+        setItemsPerPage(newRange);
+    };
+
+    const onSort = (newColumnHash: string, newDirection: TableSortDirection) => {
+        setColumnHash(newColumnHash === 'stock' ? 'inventory_level' : newColumnHash);
+        setDirection(newDirection);
+    };
 
     const renderName = (id: number, name: string): ReactElement => (
         <Link href={`/products/${id}`}>
@@ -45,15 +65,28 @@ const Products = () => {
 
     return (
         <Panel>
-            <StatefulTable
+            <Table
                 columns={[
-                    { header: 'Product name', hash: 'name', render: ({ id, name }) => renderName(id, name), sortKey: 'name' },
-                    { header: 'Stock', hash: 'stock', render: ({ stock }) => renderStock(stock), sortKey: 'stock' },
-                    { header: 'Price', hash: 'price', render: ({ price }) => renderPrice(price), sortKey: 'price' },
-                    { header: 'Action', hideHeader: true, hash: 'id', render: ({ id }) => renderAction(id), sortKey: 'id' },
+                    { header: 'Product name', hash: 'name', render: ({ id, name }) => renderName(id, name), isSortable: true },
+                    { header: 'Stock', hash: 'stock', render: ({ stock }) => renderStock(stock), isSortable: true },
+                    { header: 'Price', hash: 'price', render: ({ price }) => renderPrice(price), isSortable: true },
+                    { header: 'Action', hideHeader: true, hash: 'id', render: ({ id }) => renderAction(id) },
                 ]}
                 items={tableItems}
                 itemName="Products"
+                pagination={{
+                    currentPage,
+                    totalItems: meta?.pagination?.total,
+                    onPageChange: setCurrentPage,
+                    itemsPerPageOptions,
+                    onItemsPerPageChange,
+                    itemsPerPage,
+                }}
+                sortable={{
+                  columnHash,
+                  direction,
+                  onSort,
+                }}
                 stickyHeader
             />
         </Panel>
--- a/test/pages/products/__snapshots__/index.spec.tsx.snap
+++ b/test/pages/products/__snapshots__/index.spec.tsx.snap
@@ -17,6 +17,137 @@ exports[`Product List renders correctly 1`] = `
         5 Products
       </p>
     </div>
+    <div
+      class="styled__StyledBox-sj5f1m-0 jvmRZj styled__StyledFlexItem-smjqtt-0 cOAnic styled__StyledPaginationContainer-sc-1w0s9nk-0 iyUEaK"
+    >
+      <div
+        aria-label="pagination"
+        class="styled__StyledBox-sj5f1m-0 jvmRZj styled__StyledFlex-hcvk8l-0 kObzyk"
+        role="navigation"
+      >
+        <div
+          class="styled__StyledBox-sj5f1m-0 jvmRZj styled__StyledFlexItem-smjqtt-0 gJMDHe"
+        >
+          <div
+            class="styled__StyledBox-sj5f1m-0 jvmRZj styled__StyledBox-sc-2535c9-1 hFoBcj"
+          >
+            <button
+              aria-expanded="false"
+              aria-haspopup="listbox"
+              aria-labelledby="bd-dropdown-2-label bd-dropdown-2-toggle-button"
+              class="styled__StyledButton-sc-3yq204-0 fxqqOW styled__StyledButton-sc-12shuvi-0 bjsrwI"
+              id="bd-dropdown-2-toggle-button"
+              role="button"
+              tabindex="0"
+            >
+              <span
+                class="styled__ContentWrapper-sc-3yq204-1 cUnSLq"
+              >
+                1 - 5 of 5
+                <svg
+                  class="base__StyledIcon-a9u0e1-0 krWsIF"
+                  fill="currentColor"
+                  height="24"
+                  stroke="currentColor"
+                  stroke-width="0"
+                  viewBox="0 0 24 24"
+                  width="24"
+                >
+                  <path
+                    d="M0 0h24v24H0V0z"
+                    fill="none"
+                  />
+                  <path
+                    d="M8.71 11.71l2.59 2.59c.39.39 1.02.39 1.41 0l2.59-2.59c.63-.63.18-1.71-.71-1.71H9.41c-.89 0-1.33 1.08-.7 1.71z"
+                  />
+                </svg>
+              </span>
+            </button>
+            <ul
+              aria-labelledby="bd-dropdown-2-label"
+              class="styled__StyledList-sc-5w3dcq-0 pzdvE"
+              data-placement="bottom-start"
+              id="bd-dropdown-2-menu"
+              role="listbox"
+              style="position: absolute; left: 0px; top: 0px; transform: translate(0px, 4px);"
+              tabindex="-1"
+            />
+          </div>
+        </div>
+        <div
+          class="styled__StyledBox-sj5f1m-0 jvmRZj styled__StyledFlexItem-smjqtt-0 gJMDHe"
+        >
+          <button
+            class="styled__StyledButton-sc-3yq204-0 kDFznp styled__StyledButton-sc-12shuvi-0 bjsrwI"
+            disabled=""
+            role="button"
+            tabindex="0"
+          >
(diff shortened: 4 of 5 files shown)


test/pages/products/__snapshots__/index.spec.tsx.snap
```diff
@@ -14,9 +14,139 @@ exports[`Product List renders correctly 1`] = `
       <p
         class="styled__StyledText-tqnj75-5 hCSmtq"
       >
-        5 Products
+        undefined Products
```
bc-zachary, line 17: This is a little concerning and doesn't match what actually renders when running locally.

bc-chaz: so the reason this is coming in as undefined is because we haven't expanded the mock data to include `meta`.  If you look at the `Table` pagination prop, there's a `totalItems` that's coming from meta. `totalItems: meta?.pagination?.total`

lib/hooks.ts
```diff
@@ -30,13 +30,15 @@ export function useProducts() {
...
+export function useProductList(page?: number, limit?: number) {
     const encodedContext = useSession()?.context;
+    const query = `&page=${page}&limit=${limit}`;
     // Use an array to send multiple arguments to fetcher
-    const { data, error, mutate: mutateList } = useSWR(encodedContext ? ['/api/products/list', encodedContext] : null, fetcher);
+    const { data, error, mutate: mutateList } = useSWR(encodedContext ? [`/api/products/list`, encodedContext, query] : null, fetcher);
 
     return {
-        list: data,
+        list: data?.data,
```
bc-chaz, line 40: we don't need optional chaining for top-level props - please review documentation: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#syntax

lib/hooks.ts
```diff
@@ -30,13 +30,15 @@ export function useProducts() {
     };
 }
 
-export function useProductList() {
+export function useProductList(page?: number, limit?: number) {
     const encodedContext = useSession()?.context;
+    const query = `&page=${page}&limit=${limit}`;
     // Use an array to send multiple arguments to fetcher
-    const { data, error, mutate: mutateList } = useSWR(encodedContext ? ['/api/products/list', encodedContext] : null, fetcher);
+    const { data, error, mutate: mutateList } = useSWR(encodedContext ? [`/api/products/list`, encodedContext, query] : null, fetcher);
```
bc-chaz, line 37: `/api/products/list` should be wrapped in single quotes, not as a template literal - we're not concatenating or passing in args

pages/products/index.tsx
```diff
@@ -1,23 +1,38 @@
...
         price,
         stock,
     }));
 
+    useEffect(() => {
+        if (list.length && meta) {
+            setCurrentPage(meta.pagination?.current_page);
+            setItemsPerPage(meta.pagination?.per_page);
+        }
+    }, [currentPage, list, meta]);
```
bc-chaz, line 29: rather than passing in objects (i.e. `list` and `meta`), we should destructure and pass in only the primitive values we need. useEffect, like most of JS, uses referential equality when comparing objects (i.e. in the dependency array). Essentially, this means that passing in an object will cause this useEffect to run every time

test/pages/products/__snapshots__/index.spec.tsx.snap
bc-chaz: specifically, look at how we've handled `list` and think about how we might handle/ add `meta`: https://github.com/bigcommerce/sample-app-nodejs/blob/main/test/mocks/hooks.ts#L28

pages/products/index.tsx
```diff
@@ -1,23 +1,38 @@
...
 import { useRouter } from 'next/router';
-import { ReactElement } from 'react';
+import { ReactElement, useEffect, useState } from 'react';
 import ErrorMessage from '../../components/error';
 import Loading from '../../components/loading';
 import { useProductList } from '../../lib/hooks';
 import { TableItem } from '../../types';
 
 const Products = () => {
+    const [itemsPerPageOptions] = useState([10, 20, 50, 100]);
```
bc-chaz, line 12: this is odd - we're initializing state, but then have no method to set it, nor does it look like we ever want to... this should just be a const, sans `useState`

tylerbroadway (approved):

lib/hooks.ts
```diff
@@ -30,13 +30,17 @@ export function useProducts() {
...
+    const limitQuery = limit ? `&limit=${limit}` : '';
+    const query = `${pageQuery}${limitQuery}`;
     // Use an array to send multiple arguments to fetcher
-    const { data, error, mutate: mutateList } = useSWR(encodedContext ? ['/api/products/list', encodedContext] : null, fetcher);
+    const { data, error, mutate: mutateList } = useSWR(encodedContext ? ['/api/products/list', encodedContext, query] : null, fetcher);
 
     return {
-        list: data,
+        list: data ? data.data : [],
+        meta: data ? data.meta : {},
```
bc-chaz, line 43: I wouldn't initialize this to an empty object - this can have unintended side-effects. We can simply do `data?.meta` here. Note: in this instance we should use optional chaining on meta since `data` is in fact a destructured (nested) object - https://github.com/bigcommerce/sample-app-nodejs/blob/main/pages/api/products/list.ts#L13
```suggestion
        meta: data?.meta,
```

lib/hooks.ts
```diff
@@ -30,13 +30,17 @@ export function useProducts() {
...
+    const pageQuery = page ? `&page=${page}` : '';
+    const limitQuery = limit ? `&limit=${limit}` : '';
+    const query = `${pageQuery}${limitQuery}`;
     // Use an array to send multiple arguments to fetcher
-    const { data, error, mutate: mutateList } = useSWR(encodedContext ? ['/api/products/list', encodedContext] : null, fetcher);
+    const { data, error, mutate: mutateList } = useSWR(encodedContext ? ['/api/products/list', encodedContext, query] : null, fetcher);
 
     return {
-        list: data,
+        list: data ? data.data : [],
```
bc-chaz, line 42: same here, let's simplify:
```suggestion
        list: data.data,
```

bc-zachary: @bc-chaz I think this or something similar may actually be necessary now because of how the shape of the data has changed. Since we have two separate items now inside of the data object and need to reference them individually, we run into an issue with SWR where `data` is undefined until it is loaded. `list: data.data,` results in a runtime error `TypeError: Cannot read properties of undefined (reading 'data')`. This is what I was trying to work around with the optional chaining before and now this. 

Before, when `data` only contained the product list array, we weren't referencing anything inside of it (`list: data` as you can see in the diff) so it didn't matter that `data` is undefined during the loading state.

lib/hooks.ts
```diff
@@ -30,13 +30,17 @@ export function useProducts() {
     };
 }
 
-export function useProductList() {
+export function useProductList(page?: number, limit?: number) {
     const encodedContext = useSession()?.context;
+    const pageQuery = page ? `&page=${page}` : '';
+    const limitQuery = limit ? `&limit=${limit}` : '';
+    const query = `${pageQuery}${limitQuery}`;
```
bc-chaz, line 37: this looks like it could get a bit messy, especially if we need to add additional params.  to make this a bit more dynamic, we can make use of `URLSearchParams`. **Note**: I recommend changing the props to a props Object:

e.g.
```jsx
// Place in the ../types folder somewhere
interface QueryProps {
    [key: string]: string | number | boolean; // might be able to expand QueryParams interface
}

export function useProductList(query?: QueryProps) {
    const encodedContext = useSession()?.context;

    const queryEntries = Object.entries(query);
    const params = new URLSearchParams();
    if (queryEntries.length) {
        queryEntries.forEach(([key, value]) => params.append(key, String(value)));
    }
    // Use an array to send multiple arguments to fetcher
    const { data, error, mutate: mutateList } = useSWR(encodedContext ? ['/api/products/list', encodedContext, params.toString()] : null, fetcher);
    ...
}
```
And then you'll call the hook like `useProductList({ page: 2, limit: 10 })`

lib/hooks.ts
```diff
@@ -2,8 +2,8 @@ import useSWR from 'swr';
 import { useSession } from '../context/session';
 import { ErrorProps, ListItem } from '../types';
 
-async function fetcher(url: string, encodedContext: string) {
-    const res = await fetch(`${url}?context=${encodedContext}`);
+async function fetcher(url: string, encodedContext: string, query = '') {
+    const res = await fetch(`${url}?context=${encodedContext}${query}`);
```
bc-chaz, line 6: once we do the change below, I believe we'll need to add a `&` before query here

lib/hooks.ts
bc-chaz: now that I'm looking at this file, I'm realizing that we're passing `encodedContext` as a param - this might actually be a good time to just make the other hooks more URL param friendly. We can actually modify the sample code I gave above such that we send only one param (query string) to the fetcher function:
```jsx
async function fetcher(url: string, query: string) {
    const res = await fetch(`${url}?${query}`);
    ...
}

export function useProducts() {
    const encodedContext = useSession()?.context;
    const params = new URLSearchParams();
    params.append('context', encodedContext);
    // Request is deduped and cached; Can be shared across components
    const { data, error } = useSWR(encodedContext ? ['/api/products', params.toString()] : null, fetcher);

    return {
        summary: data,
        isLoading: !data && !error,
        error,
    };
}

export function useProductList(query?: QueryProps) {
    const encodedContext = useSession()?.context;
    const params = new URLSearchParams();
    params.append('context', encodedContext);

    const queryEntries = Object.entries(query);
    if (queryEntries.length) {
        queryEntries.forEach(([key, value]) => params.append(key, String(value)));
    }
    // Use an array to send multiple arguments to fetcher
    const { data, error, mutate: mutateList } = useSWR(encodedContext ? ['/api/products/list', params.toString()] : null, fetcher);
    ...
}
```

pages/api/products/list.ts
```diff
@@ -5,13 +5,10 @@ export default async function list(req: NextApiRequest, res: NextApiResponse) {
...
         const bigcommerce = bigcommerceClient(accessToken, storeHash);
-        // Optional: pass in API params here
-        const params = [
-            'limit=11',
-        ].join('&');
+        const { page, limit } = req.query;
 
-        const { data } = await bigcommerce.get(`/catalog/products?${params}`);
-        res.status(200).json(data);
+        const response = await bigcommerce.get(`/catalog/products?page=${page}&limit=${limit}`);
```
bc-chaz, line 10: this is probably good for now. If in the future we want to expand the params, we can also make use of URLSearchParams:
```jsx
const params = new URLSearchParams({ limit, page, search, somethingElse }).toString();
const response = await bigcommerce.get(`/catalog/products?${params}`);
```

lib/hooks.ts
bc-chaz: err so currently, `data` is a destructure of the response - https://github.com/bigcommerce/sample-app-nodejs/blob/main/pages/api/products/list.ts#L13. SWR might be doing something internal to account for this. If we're getting an error here, I'd recommend adding in optional chaining and allowing an undefined state:
```jsx
return {
    list: data?.data,
    meta: data?. meta,
}
```

pages/products/index.tsx
```diff
@@ -45,15 +53,23 @@ const Products = () => {
...
+                    { header: 'Product name', hash: 'name', render: ({ id, name }) => renderName(id, name), isSortable: true },
+                    { header: 'Stock', hash: 'stock', render: ({ stock }) => renderStock(stock), isSortable: true },
+                    { header: 'Price', hash: 'price', render: ({ price }) => renderPrice(price), isSortable: true },
+                    { header: 'Action', hideHeader: true, hash: 'id', render: ({ id }) => renderAction(id), isSortable: true },
                 ]}
                 items={tableItems}
                 itemName="Products"
+                pagination={{
+                    currentPage,
+                    totalItems: meta.pagination?.total,
```
bc-chaz, line 67: since meta itself is a nested prop, we should add optional chaining to pagaination:
```suggestion
                    totalItems: meta?.pagination?.total,
```

lib/hooks.ts
```diff
@@ -30,24 +31,28 @@ export function useProducts() {
     };
 }
 
-export function useProductList() {
-    const encodedContext = useSession()?.context;
+export function useProductList(query?: QueryProps) {
+    const { context } = useSession();
+    const params = new URLSearchParams({ ...query, context }).toString(); // Initialize to incoming query
```
bc-chaz, line 36: let's get rid of the comment here - that was just for the sample code

types/auth.ts
```diff
@@ -24,3 +24,7 @@ export interface SessionContextProps {
 export interface QueryParams {
     [key: string]: string | string[];
 }
+
+export interface QueryProps {
+    [key: string]: string; // might be able to expand QueryParams interface
+}
```
bc-chaz, line 30: I believe GH is upset that there's not an empty line at the end of this file - there might be config in your IDE to automatically add this on save

types/auth.ts
```diff
@@ -24,3 +24,7 @@ export interface SessionContextProps {
 export interface QueryParams {
     [key: string]: string | string[];
 }
+
+export interface QueryProps {
+    [key: string]: string; // might be able to expand QueryParams interface
```
bc-chaz, line 29: same with the comment here - this was only to explain the code in the comments. Also, this now looks largely similar to `QueryParams` - we can probably remove this and just point to `QueryParams` in hooks

pages/products/index.tsx
```diff
@@ -45,15 +53,23 @@ const Products = () => {
...
+            <Table
                 columns={[
-                    { header: 'Product name', hash: 'name', render: ({ id, name }) => renderName(id, name), sortKey: 'name' },
-                    { header: 'Stock', hash: 'stock', render: ({ stock }) => renderStock(stock), sortKey: 'stock' },
-                    { header: 'Price', hash: 'price', render: ({ price }) => renderPrice(price), sortKey: 'price' },
-                    { header: 'Action', hideHeader: true, hash: 'id', render: ({ id }) => renderAction(id), sortKey: 'id' },
+                    { header: 'Product name', hash: 'name', render: ({ id, name }) => renderName(id, name), isSortable: true },
+                    { header: 'Stock', hash: 'stock', render: ({ stock }) => renderStock(stock), isSortable: true },
+                    { header: 'Price', hash: 'price', render: ({ price }) => renderPrice(price), isSortable: true },
+                    { header: 'Action', hideHeader: true, hash: 'id', render: ({ id }) => renderAction(id), isSortable: true },
```
bc-chaz, line 61: in order to properly sort, we need to add an additional top-level, prop to `Table`, `sortable`. We'll also need to add some additional state in order for the Table to properly handle sorting. Please see `Sortable` -https://bigcommerce.github.io/big-design/table
```jsx
const router = useRouter();
const [tableItems, setTableItems] = useState([]);
const [itemsPerPage, setItemsPerPage] = useState(10);
const [currentPage, setCurrentPage] = useState(1);
const [columnHash, setColumnHash] = useState('');
const [direction, setDirection] = useState<'ASC' | 'DESC'>('ASC');
const { error, isLoading, list = [], meta = {} } = useProductList({ page: String(currentPage), limit: String(itemsPerPage) });

useEffect(() => {
    if (list.length) {
        const tableItems: TableItem[] = list.map(({ id, inventory_level: stock, name, price }) => ({
            id,
            name,
            price,
            stock,
        }));
        setTableItems(tableItems);
    }
}, [list]);

const onSort = (newColumnHash: string, newDirection: string) => {
    setColumnHash(newColumnHash);
    setDirection(newDirection);
    setTableItems(currentItems => sort(currentItems, newColumnHash, newDirection));
};

return (
    <Table
        columns={[...]}
        pagination={{ ... }}
        sortable={{
            columnHash,
            direction,
            onSort,
        }}
    />
);
```

bc-chaz: since the `useEffect` above is passing an array (`list`) as a dependency, we'll want to memoize that array. Most of JS uses referential equality when comparing two objects - since arrays are a type of object, `arr1 === arr2` will always be false, even if they contain exactly the same key/ values. The only time arr1 will equal arr2 is if we explicitly set `arr1 = arr2`.

In our case, this means we'll need to update the hook so that this useEffect doesn't get hit on every lifecycle. In our `lib/hooks` we might be able to do a quick n dirty useMemo as part of the return:
```jsx
return useMemo(() => ({
    list: data?.data,
    meta: data?.meta,
    isLoading: !data && !error,
    error,
    mutateList,
}, [data, error, mutateList]));
```

pages/products/index.tsx
```diff
@@ -1,23 +1,43 @@
...
+    const [currentPage, setCurrentPage] = useState(1);
+    const [columnHash, setColumnHash] = useState('id');
+    const [sortDirection, setSortDirection] = useState<TableSortDirection>('ASC');
     const router = useRouter();
-    const { error, isLoading, list = [] } = useProductList();
+    const { error, isLoading, list = [], meta = {} } = useProductList({
+      page: String(currentPage),
+      limit: String(itemsPerPage),
+      sort: columnHash,
+      direction: sortDirection === 'ASC' ? 'asc' : 'desc',
```
bc-zachary, line 21: I'm converting to lowercase this way because `Table`'s direction attribute needs to be of type `TableSortDirection` which prevents me from using `toLowercase()` without first changing it to type string. This felt like a cleaner option.

pages/api/products/list.ts
```diff
@@ -5,13 +5,12 @@ export default async function list(req: NextApiRequest, res: NextApiResponse) {
...
-        // Optional: pass in API params here
-        const params = [
-            'limit=11',
-        ].join('&');
+        const { page, limit, sort, direction } = req.query;
 
-        const { data } = await bigcommerce.get(`/catalog/products?${params}`);
-        res.status(200).json(data);
+        const response = await bigcommerce.get(
+          `/catalog/products?page=${page}&limit=${limit}&sort=${sort}&direction=${direction}`
```
bc-chaz, line 11: since the params have expanded, we should probably make use of `URLSearchParams` - even more so given that sort and direction should be optional

pages/products/index.tsx
```diff
@@ -1,23 +1,43 @@
...
+import { ReactElement, useState } from 'react';
 import ErrorMessage from '../../components/error';
 import Loading from '../../components/loading';
 import { useProductList } from '../../lib/hooks';
 import { TableItem } from '../../types';
 
 const Products = () => {
+    const [itemsPerPage, setItemsPerPage] = useState(10);
+    const [currentPage, setCurrentPage] = useState(1);
+    const [columnHash, setColumnHash] = useState('id');
```
bc-chaz, line 14: this should be an empty string - we want to have the default sort unless specified by the user

pages/products/index.tsx
```diff
@@ -1,23 +1,43 @@
...
+    const [sortDirection, setSortDirection] = useState<TableSortDirection>('ASC');
     const router = useRouter();
-    const { error, isLoading, list = [] } = useProductList();
+    const { error, isLoading, list = [], meta = {} } = useProductList({
+      page: String(currentPage),
+      limit: String(itemsPerPage),
+      sort: columnHash,
```
bc-chaz, line 20: let's only set sort and direction if the user specifies a sort... we should be able to do a conditional add of a property here.
e.g.
```jsx
{
    keyOne: 'some value',
    ...(conditionOrVarToCheck && { keyTwo: 'some other value' }),
}
```

pages/products/index.tsx
```diff
@@ -45,15 +65,28 @@ const Products = () => {
...
+                    currentPage,
+                    totalItems: meta?.pagination?.total,
+                    onPageChange: (newPage => setCurrentPage(newPage)),
+                    itemsPerPageOptions,
+                    onItemsPerPageChange,
+                    itemsPerPage,
+                }}
+                sortable={{
+                  columnHash,
+                  direction: sortDirection,
```
bc-chaz, line 87: why not just call `sortDirection`, direction?

pages/products/index.tsx
bc-chaz: I believe we can still use `.toLowerCase()` here. We probably don't want to do a manual conversion of strings unless we absolutely have to.
`direction: direction.toLowerCase()`

pages/products/index.tsx
```diff
@@ -45,15 +65,28 @@ const Products = () => {
...
+            <Table
                 columns={[
-                    { header: 'Product name', hash: 'name', render: ({ id, name }) => renderName(id, name), sortKey: 'name' },
-                    { header: 'Stock', hash: 'stock', render: ({ stock }) => renderStock(stock), sortKey: 'stock' },
-                    { header: 'Price', hash: 'price', render: ({ price }) => renderPrice(price), sortKey: 'price' },
-                    { header: 'Action', hideHeader: true, hash: 'id', render: ({ id }) => renderAction(id), sortKey: 'id' },
+                    { header: 'Product name', hash: 'name', render: ({ id, name }) => renderName(id, name), isSortable: true },
+                    { header: 'Stock', hash: 'stock', render: ({ stock }) => renderStock(stock), isSortable: true },
+                    { header: 'Price', hash: 'price', render: ({ price }) => renderPrice(price), isSortable: true },
+                    { header: 'Action', hideHeader: true, hash: 'id', render: ({ id }) => renderAction(id), isSortable: true },
```
bc-chaz, line 73: let's get rid of `isSortable` for the action column... there's this weird effect whereby the current implementation is causing the action column to be sortable by default and it shows a sort arrow for the column:
Screen Shot 2021-09-24 at 12 09 08 PM
Since the column contains only a 3 dot menu (`...`), sorting a column that shows same exact content seems odd.

pages/api/products/list.ts
bc-chaz: e.g.
```jsx
const { page, limit, sort, direction } = req.query;
const params = new URLSearchParams({ page, limit, ...(sort && { sort, direction }) }).toString();

const response = await bigcommerce.get(`/catalog/products?${params}`);
```

lib/hooks.ts
```diff
@@ -30,24 +31,28 @@ export function useProducts() {
...
         isLoading: !data && !error,
         error,
         mutateList,
     };
 }
 
 export function useProductInfo(pid: number, list: ListItem[]) {
-    const encodedContext = useSession()?.context;
+    const { context } = useSession();
+    const params = new URLSearchParams({ context: context }).toString();
```
bc-chaz, line 52: if the object property name is equal to a variable, we should use the [shorthand property name](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#new_notations_in_ecmascript_2015). DRY
```suggestion
    const params = new URLSearchParams({ context }).toString();
```

bc-chaz (approved):
👍 LGTM

PR #68, opened 2021-11-04
fix(common): persistant session across pages

chanceaclark:
## What?
Creates a better persistence of session across all pages.

## Why?
If we end up loading the app on the `/products` page, we'll get the context from the load URL and pass it into the session context.

### DIFF
--- a/context/session.tsx
+++ b/context/session.tsx
@@ -1,14 +1,23 @@
-import { createContext, useContext, useState } from 'react';
-import { ContextValues } from '../types';
+import { useRouter } from 'next/router';
+import { createContext, useContext, useEffect, useState } from 'react';
+import { bigCommerceSDK } from '../scripts/bcSdk';
 
-const SessionContext = createContext<Partial<ContextValues>>({});
+const SessionContext = createContext({ context: '' });
 
 const SessionProvider = ({ children }) => {
+    const { query } = useRouter();
     const [context, setContext] = useState('');
-    const value = { context, setContext };
+
+    useEffect(() => {
+        if (query.context) {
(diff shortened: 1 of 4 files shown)


jorgemoya (approved):
Thanks for updating this. 👍

pages/index.tsx
```diff
@@ -1,18 +1,12 @@
 import { Box, Flex, H1, H4, Panel } from '@bigcommerce/big-design';
-import { useEffect } from 'react';
+import React from 'react';
```
bc-chaz, line 2: are we using the `React` variable anywhere? If not, I believe Next will automatically import this for us

context/session.tsx
```diff
@@ -1,14 +1,23 @@
...
     const [context, setContext] = useState('');
-    const value = { context, setContext };
+
+    useEffect(() => {
+        if (query.context) {
+            setContext(query.context.toString());
+            // Keeps app in sync with BC (e.g. heatbeat, user logout, etc)
+            bigCommerceSDK(query.context);
+        }
+    }, [query.context]);
```
bc-chaz, line 17: 🎉

bc-chaz:
one minor comment, otherwise LGTM. always love it when we remove more lines than add 👍

pages/index.tsx
chanceaclark: Ah that's a habit, forgot that's apart of how [React transforms JSX](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html)

bc-chaz (approved):

PR #69, opened 2021-11-05
feat(common): use jwt url value for load redirect

chanceaclark:
## What?
Use the JWT token `url` key/value to load redirect URL.

## Why?
This allows paths to be provided when the app loads. For instance, we want to load `/app/6/orders/100?foo=1&bar=2` the app will now be able to redirect to `/orders/100?context=<encodedContext>&foo=1&bar=2`.

## Testing / Proof
Params passed into the app:
Screen Shot 2021-11-05 at 11 27 15

### DIFF
--- a/pages/api/load.ts
+++ b/pages/api/load.ts
@@ -1,6 +1,13 @@
 import { NextApiRequest, NextApiResponse } from 'next';
 import { encodePayload, getBCVerify, setSession } from '../../lib/auth';
 
+const buildRedirectUrl = (url: string, encodedContext: string) => {
+    const [path, query = ''] = url.split('?');
+    const queryParams = new URLSearchParams(`context=${encodedContext}&${query}`);
+
+    return `${path}?${queryParams}`;
+}
+
 export default async function load(req: NextApiRequest, res: NextApiResponse) {
     try {
         // Verify when app loaded (launch)
@@ -8,7 +15,7 @@ export default async function load(req: NextApiRequest, res: NextApiResponse) {
         const encodedContext = encodePayload(session); // Signed JWT to validate/ prevent tampering
 
         await setSession(session);
-        res.redirect(302, `/?context=${encodedContext}`);
+        res.redirect(302, buildRedirectUrl(session.url, encodedContext));
     } catch (error) {
         const { message, response } = error;
         res.status(response?.status || 500).json({ message });


jorgemoya (approved):
:gucci:

pages/api/load.ts
```diff
@@ -1,14 +1,21 @@
...
+
 export default async function load(req: NextApiRequest, res: NextApiResponse) {
     try {
         // Verify when app loaded (launch)
         const session = await getBCVerify(req.query);
         const encodedContext = encodePayload(session); // Signed JWT to validate/ prevent tampering
 
         await setSession(session);
-        res.redirect(302, `/?context=${encodedContext}`);
+        res.redirect(302, buildRedirectUrl(session.url, encodedContext));
```
bc-chaz, line 18: 🍹 I like the idea of placing this bit of logic in either a const or helper function. I do wonder, however, if we're overcomplicating things a bit. In essence, we're taking the `session.url`, splitting that into an array of `path` and `query`, then taking the `query` to pass into `URLSearchParams`, only to recombine that in a template literal with the path. Would it not be simpler to do `${session.url}?context=${encodedContext}` since that's in effect what we're doing here?

Usually, I'll use `URLSearchParams` if I want to convert an object into a param string or if I want to compute/ derive some data (e.g. get, build, or remove a param). Here, we're calling the urlSearchParams constructor only to immediately convert back into a string, without making use of anything that `URLSearchParams` has to offer.

chanceaclark: @bc-chaz The reasoning behind not doing it that way is if we pass query params into `session.url` is that it'll build a malformed url (e.g. `/orders/100?foo=1&bar=2?context=<encodedContext>` (notice the second `?`).

I used `URLSearchParams` to validate that the params coming after the path are actually valid and to strip out anything like hashes.

bc-chaz: yee, I think what struck me as odd is that the input to the params constructor is exactly the same as the output we're looking for.  
i.e.
```jsx
const queryParams = new URLSearchParams(`${query}&context=${encodedContext}`);
const params = `${query}&context=${encodedContext}`;
```
Both of these are exactly the same (sans encoding) when we convert to a string (either via .toString() or template literal). In terms of encoding characters, would we want to convert hashes since those are usually anchor tags? If we're concerned about non-URI specific chars, should we instead use something like `encodeURI`?

chanceaclark: So if I do this, it's a little wonky,
```tsx
const query = ''; // assume no params were passed
const params = `${query}&context=${encodedContext}`;

return `${path}?${params}`; // '/orders/100?&context=<encodedContext>'
// Still valid but pretty wonky
```

The goal was to build a clean URL with as little of manual effort as possible.

I'm a little confused about the last part, mostly just not sure what to make of it quite yet. I wasn't specifically worried about encoding at this point. Definitely should be looked at, but maybe for another round?

bc-chaz: actually, now that I'm looking at this, I think this will break an anchor tag in either case.
e.g.
```jsx
const url = '/orders?foo=1&bar=2#somewhere';
const [path, query = ''] = url.split('?');
const queryParams = new URLSearchParams(`${query}&context=${encodedContext}`);

// foo=1&bar=2%23somewhere&context=ey2.fhtr
```
If we use `encodeURI`, the hash won't be converted, but the anchor will still be smushed between the queries... we should probably list the `context` first and then add the query/ rest of the uri after.

chanceaclark: We had a convo offline, this is not a blocking comment. We also discussed that we don't pass a hash in so it doesn't matter right now, but we should be pull the context query param before the passed in query.

PR #70, opened 2021-11-15
feat(common): add ThemeProvider to _app.tsx

chanceaclark:
## What?
Adds a `ThemeProvider` to the application.

## Why?
Provides a way to use the theme when building styled-components.

For example,

```tsx
import styled from 'styled-components';

const StyledDl = styled.dl`
  color: ${({ theme }) => theme.colors.secondary70};
  line-height: ${({ theme }) => theme.lineHeight.medium};
`;
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -802,6 +802,17 @@
         "react-datepicker": "^2.16.0",
         "react-popper": "^2.2.3",
         "react-uid": "^2.2.0"
+      },
+      "dependencies": {
+        "@bigcommerce/big-design-theme": {
+          "version": "0.11.0",
+          "resolved": "https://registry.npmjs.org/@bigcommerce/big-design-theme/-/big-design-theme-0.11.0.tgz",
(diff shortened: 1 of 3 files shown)


bc-zachary (approved):
LGTM 👍

bc-chaz:
@chanceaclark I'm a bit torn on this PR. Believe it or not, it was actually a deliberate effort to not include a theme provider with the sample app.  The rationale was that design (via Sophi) really wanted to discourage the use of styled components in favor of using BigDesign almost exclusively - to the point where we actually had to redesign/ engineer the product form to use plain BigDesign components. Personally, I think it would have been a lot easier had we included a provider.

I realize, of course, that the new orders page relies on the BD theme.  The issue is by doing this we're in effect showing devs how to deviate from BD in the main sample app, which design was very much so against. Can we discuss with Booker and design to see if we're OK changing course?

jorgemoya (approved):

chanceaclark:
Offline discussion, we decided to move forward with this.

bc-chaz (approved):

PR #71, opened 2021-11-17
feat(common): add order page for modal

chanceaclark:
## What?
Builds off of #69 & #70.

Adds a modal view for viewing orders.

## Why?
Primarily builds off of #69 to support deep linking within modals. This is an example view we might ship out.

### DIFF
--- a/components/error.tsx
+++ b/components/error.tsx
@@ -1,11 +1,23 @@
 import { H3, Panel } from '@bigcommerce/big-design';
-import { ErrorMessageProps } from '../types';
+import { ErrorMessageProps, ErrorProps } from '../types';
 
-const ErrorMessage = ({ error }: ErrorMessageProps) => (
-    <Panel>
+const ErrorContent = ({ message }: Pick<ErrorProps, 'message'>) => (
+    <>
         <H3>Failed to load</H3>
-        {error && error.message}
-    </Panel>
-);
+        {message}
+    </>
+)
+
+const ErrorMessage = ({ error, renderPanel = true }: ErrorMessageProps) => {
+    if (renderPanel) {
+        return (
+            <Panel>
+                <ErrorContent message={error.message} />
+            </Panel>
+        )
+    }
+
+    return <ErrorContent message={error.message} />
+};
 
 export default ErrorMessage;
--- a/components/header.tsx
+++ b/components/header.tsx
@@ -13,6 +13,11 @@ export const TabRoutes = {
     [TabIds.PRODUCTS]: '/products',
 };
 
+const HeaderlessRoutes = [
+    '/orders/[orderId]/labels',
+    '/orders/[orderId]/modal',
+]
+
 const InnerRoutes = [
     '/products/[pid]',
 ];
@@ -20,6 +25,7 @@ const InnerRoutes = [
 const HeaderTypes = {
     GLOBAL: 'global',
     INNER: 'inner',
+    HEADERLESS: 'headerless',
 };
 
 const Header = () => {
@@ -32,6 +38,8 @@ const Header = () => {
         if (InnerRoutes.includes(pathname)) {
             // Use InnerHeader if route matches inner routes
             setHeaderType(HeaderTypes.INNER);
+        } else if (HeaderlessRoutes.includes(pathname)) {
+            setHeaderType(HeaderTypes.HEADERLESS);
         } else {
             // Check if new route matches TabRoutes
             const tabKey = Object.keys(TabRoutes).find(key => TabRoutes[key] === pathname);
@@ -59,6 +67,7 @@ const Header = () => {
         return router.push(TabRoutes[tabId]);
     };
 
+    if (headerType === HeaderTypes.HEADERLESS) return null;
     if (headerType === HeaderTypes.INNER) return <InnerHeader />;
 
     return (
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -9,7 +9,10 @@ const MyApp = ({ Component, pageProps }: AppProps) => {
     return (
         <ThemeProvider theme={defaultTheme}>
             <GlobalStyles />
-            <Box marginHorizontal="xxxLarge" marginVertical="xxLarge">
+            <Box
+                marginHorizontal={{ mobile: 'none', tablet: 'xxxLarge' }}
+                marginVertical={{ mobile: 'none', tablet: "xxLarge" }}
+            >
                 <Header />
                 <SessionProvider>
                     <Component {...pageProps} />
--- a/pages/orders/[orderId]/modal.tsx
+++ b/pages/orders/[orderId]/modal.tsx
@@ -1,0 +1,121 @@
+import { Badge, Flex, Grid, GridItem, H3, Text } from '@bigcommerce/big-design';
+import { useRouter } from 'next/router';
+import React from 'react';
+import styled from 'styled-components';
+
+import ErrorMessage from '../../../components/error';
+import { useOrder } from '../../../lib/hooks';
+import { Order } from '../../../types';
+
+const StyledAddress = styled.address`
+    color: ${({ theme }) => theme.colors.secondary70};
+    font-style: normal;
+    line-height: ${({ theme }) => theme.lineHeight.medium};
+`;
+
+const StyledDl = styled.dl`
+    color: ${({ theme }) => theme.colors.secondary70};
+    display: grid;
+    grid-template: 'dt dd' auto / 1fr auto;
+    grid-auto-rows: auto;
+    line-height: ${({ theme }) => theme.lineHeight.medium};
+    margin: 0;
+
+    dt {
+        grid-area: 'dt';
+    }
+
+    dd {
+        grid-area: 'dd';
+        text-align: right;
+    }
+`;
+
+const InternalOrderModalPage = (order: Order) => {
+    const { billing_address } = order;
+
+    const formatCurrency = (amount: string) =>
+        new Intl.NumberFormat(order.customer_locale, { style: 'currency', currency: order.currency_code }).format(parseFloat(amount));
+
+    return (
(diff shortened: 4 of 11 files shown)


components/header.tsx
```diff
@@ -32,6 +38,8 @@ const Header = () => {
         if (InnerRoutes.includes(pathname)) {
             // Use InnerHeader if route matches inner routes
             setHeaderType(HeaderTypes.INNER);
+        } else if (HeaderlessRoutes.includes(pathname)) {
+            setHeaderType(HeaderTypes.HEADERLESS);
```
chanceaclark, line 42: I'm going with the existing pattern here, but I recommend we include layout components on a per-page component rather than a global layout. It'll add a little duplication here and there, but will be better for extensibility. May tackle that cleanup in a future PR.

pages/_app.tsx
```diff
@@ -1,19 +1,25 @@
...
 import type { AppProps } from 'next/app';
+import { ThemeProvider } from 'styled-components';
+
 import Header from '../components/header';
 import SessionProvider from '../context/session';
 
 const MyApp = ({ Component, pageProps }: AppProps) => {
     return (
-        <>
+        <ThemeProvider theme={defaultTheme}>
```
chanceaclark, line 11: Adding the `ThemeProvider` builds off of #70 however, the commit is not on main yet.

pages/orders/[orderId]/labels.tsx
```diff
@@ -0,0 +1,98 @@
...
+            />
+          </FormGroup>
+        </Form>
+      </GridItem>
+      <Button>Print label</Button>
+    </StyledGrid>
+  );
+};
+
+export default OrderLabelPage;
```
Comment (chanceaclark) [inline pages/orders/[orderId]/labels.tsx:98]: TBH, I dislike this page and may consider removing before merging. I primarily added this showcase a smaller modal.

pages/orders/[orderId]/labels.tsx
```diff
@@ -0,0 +1,98 @@
...
+const StyledGrid = styled(Grid)`
+  grid-template: 'form print' auto / 1fr auto;
+
+  > .bd-button {
+    position: sticky;
+    top: 0;
+  }
+`;
+
+const OrderLabelPage: React.FC = () => {
```
Comment (bc-chaz) [inline pages/orders/[orderId]/labels.tsx:25]: note: the `FC` here is unnecessary. Facebook removed this typecasting a while ago from their both their documentation and official typescript example. Also, FWIW, we're also not doing this on any other page. Please see https://github.com/facebook/create-react-app/pull/8177

pages/orders/[orderId]/labels.tsx
```diff
@@ -0,0 +1,98 @@
...
+  const [shipping, setShipping] = useState('other');
+  const [shippingType, setShippingType] = useState<string>();
+  const [checked, setChecked] = useState(false);
+
+  const toggleChecked = () => {
+    setChecked(!checked);
```
Comment (bc-chaz) [inline pages/orders/[orderId]/labels.tsx:31]: if the next state depends on the previous, we should prefer to use functional setState, in a similar fashion to setting functional updates in classes.  

Please see:
https://reactjs.org/docs/hooks-reference.html#functional-updates
https://stackoverflow.com/questions/55510565/do-we-still-need-functional-setstate-way-in-react-hooks#answer-55510736

```suggestion
    setChecked(prevChecked => !prevChecked);
```

pages/orders/[orderId]/modal.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const router = useRouter();
+  const { orderId } = router.query;
+  const { isLoading, order, error } = useOrder(parseInt(`${orderId}`, 10));
+
+  if (isLoading) {
+    return null;
+  }
+
+  if (error) {
+    return <>{error.message}</>;
```
Comment (bc-chaz) [inline pages/orders/[orderId]/modal.tsx:118]: could we reuse the error component here... or modify the error component to accept this message?
https://github.com/bigcommerce/sample-app-nodejs/blob/main/components/error.tsx

pages/orders/[orderId]/modal.tsx
```diff
@@ -0,0 +1,124 @@
...
+    </Grid>
+  );
+};
+
+const OrderModalPage: React.FC = () => {
+  const router = useRouter();
+  const { orderId } = router.query;
+  const { isLoading, order, error } = useOrder(parseInt(`${orderId}`, 10));
+
+  if (isLoading) {
```
Comment (bc-chaz) [inline pages/orders/[orderId]/modal.tsx:113]: 🍹 totally optional, but mention for consistency - on our other pages we're using a bit of modern JS to place this on one line.

e.g.
```jsx
if (isLoading) return null;
```

pages/orders/[orderId]/labels.tsx
```diff
@@ -0,0 +1,98 @@
+import {
+  Button,
```
Comment (bc-chaz) [inline pages/orders/[orderId]/labels.tsx:2]: I myself am a fan of 2 space indenting - we're unfortunately using 4 space intending in this repo. And yes, we should probably add linting for this as to not have these silly comments 😑

pages/orders/[orderId]/labels.tsx
Comment (jorgemoya) [reply pages/orders/[orderId]/labels.tsx:31]: This seems more complicated to me lol.

pages/orders/[orderId]/modal.tsx
Comment (chanceaclark) [reply pages/orders/[orderId]/modal.tsx:118]: The problem the with the error component is that it renders inside a panel. We could add a prop to conditionally render the panel, but I'm not sold it would add much value 🤷

pages/orders/[orderId]/modal.tsx
Comment (chanceaclark) [reply pages/orders/[orderId]/modal.tsx:113]: IMO, not a fan 😅 This plays into your comment below, but we should add eslint if we want consistency like this. Most of these files I've copied into a repo using our eslint config and pasted them back here.

pages/orders/[orderId]/labels.tsx
Comment (chanceaclark) [reply pages/orders/[orderId]/labels.tsx:25]: habit 😅

pages/orders/[orderId]/labels.tsx
Comment (chanceaclark) [reply pages/orders/[orderId]/labels.tsx]: I got rid of this page... it added no value to the PR.

Comment (bc-chaz) [reply pages/orders/[orderId]/labels.tsx]: @jorgemoya it has to do with how React batches calls to setState, which unfortunately React has done since class based components.  In other words, React doesn't handle multiple calls to state synchronously - the best way to guarantee future state that depends on the previous state (e.g. a counter or toggle) is to use functional updates. The stackoverflow link gives a pretty good example.

basically:
```jsx
const [count, setCount] = useState(1);

const handleClick = () => {
  setCount(count + 1);
  setCount(count + 1);
  setCount(count + 1);
}; 
// results in count === 2
```
This could also happen if the handleClick had just one `setCount`, but handleClick was hit multiple times.

components/error.tsx
```diff
@@ -1,11 +1,22 @@
...
+        )
+    }
+
+    return (
+        <>
+            <H3>Failed to load</H3>
+            {error && error.message}
+        </>
+    )
+};
```
bc-chaz, line 20: it seems a bit redundant that we're repeating the same content (H3, error message) for both returns... we could probably refactor slightly:

```suggestion
const ErrorMessage = ({ error: { message }, renderPanel = true }: ErrorMessageProps) => {
    const errorContent = (
        <>
            <H3>Failed to load</H3>
            {message && message}
        </>
    );

    if (renderPanel) return <Panel>{errorContent}</Panel>;

    return errorContent;
};
```

chanceaclark: Good point, however I did it a little differently.

components/error.tsx
```diff
@@ -1,11 +1,23 @@
...
         <H3>Failed to load</H3>
-        {error && error.message}
-    </Panel>
-);
+        {message}
+    </>
+)
+
+const ErrorMessage = ({ error, renderPanel = true }: ErrorMessageProps) => {
+    if (renderPanel) {
```
chanceaclark, line 12: I sort of don't see the point of this boolean if we could just export the `ErrorContent` component above 🤔

bc-chaz: yeah, true... if we refactor, we could probably change `ErrorContent` to `ErrorMessage` and then `ErrorMessage` becomes `WrappedErrorMessage` or something

components/error.tsx
```diff
@@ -1,11 +1,23 @@
...
+        return (
+            <Panel>
+                <ErrorContent message={error?.message} />
```
bc-chaz, line 15: given that error will always be defined, do we need optional chaining? I believe usually we do optional chaining if it's beyond one level of nesting, e.g. `parent.child?.nestedChild`

pls see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#syntax

chanceaclark: Correct 🤦 For some reason that I can't recall, I saw an undefined somewhere and added it.

components/error.tsx
chanceaclark: Let's tackle it in another PR. It'll touch a lot more points than this PR intended on touching.

bc-chaz (approved):
👍

components/error.tsx
bc-chaz: ah yeah, I didn't take your original comment as something that needed to be done for this PR - more so if we did go down that route, we might want to update the component names

chanceaclark:
Added `do not merge`. Waiting on #69.

pages/orders/[orderId]/modal.tsx
Comment (icatalina) [reply pages/orders/[orderId]/modal.tsx]: This isn't "modern" though, this pattern has existed for as long as JS. It was pretty popular in the beginning of JS (and other languages). It is not recommended though, it goes against our code style guide and it can lead to issues. It also makes the diff worse when needing to add new stuff to the if block.

```diff
-if (isLoading) return null;
+if (isLoading) {
+  console.log('something else');
+  return null;
+}
```

vs:

```diff
if (isLoading) {
+  console.log('something else');
  return null;
}
```

PR #72, opened 2021-11-18
feat(common): updates npm, packages, and tests

bc-chaz:
## What?
Updates sample app to use NPM 7+. This effectively creates a new package-lock file using `lockfileVersion` 2. This PR also updates big-design, NextJS (fixes security vulnerability), and styled-components. Unit tests/ snapshots also updated.

## Why?
Addresses security vulnerabilities and ensures compatibility with workspaces.

## Testing / Proof
verified by running the app, linting, unit tests, and creating a new build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ This demo includes all of the files necessary to get started with a basic, hello
 
 To get the app running locally, follow these instructions:
 
-1. [Use Node 10+](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm#checking-your-version-of-npm-and-node-js)
+1. [Use Node 10+ and NPM 7+](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm#checking-your-version-of-npm-and-node-js)
 2. Install npm packages
     - `npm install`
 3. [Add and start ngrok.](https://www.npmjs.com/package/ngrok#usage) Note: use port 3000 to match Next's server.
--- a/package.json
+++ b/package.json
@@ -14,17 +14,21 @@
   "keywords": [],
   "author": "",
   "license": "ISC",
+  "engines": {
+    "node": ">=10.13.0 <17",
+    "npm": ">=7.20.x <8.1.4"
+  },
   "dependencies": {
-    "@bigcommerce/big-design": "^0.27.0",
+    "@bigcommerce/big-design": "^0.30.2",
     "@bigcommerce/big-design-theme": "^0.13.1",
     "firebase": "^8.2.6",
     "jsonwebtoken": "^8.5.1",
     "mysql": "^2.18.1",
-    "next": "^10.2.3",
-    "node-bigcommerce": "bigcommerce/node-bigcommerce",
+    "next": "^11.1.1",
+    "node-bigcommerce": "github:bigcommerce/node-bigcommerce",
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
-    "styled-components": "^4.4.1",
+    "styled-components": "^5.3.3",
     "swr": "^0.4.1"
   },
   "devDependencies": {
@@ -33,7 +37,7 @@
     "@types/jest": "^26.0.20",
     "@types/node": "^14.14.22",
     "@types/react": "^17.0.0",
-    "@types/styled-components": "^5.1.7",
+    "@types/styled-components": "^5.1.15",
     "@typescript-eslint/eslint-plugin": "^4.16.1",
     "@typescript-eslint/parser": "^4.16.1",
     "babel-jest": "^26.6.3",
--- a/test/pages/products/index.spec.tsx
+++ b/test/pages/products/index.spec.tsx
@@ -10,12 +10,16 @@ describe('Product List', () => {
         // Wait for table to be rendered
         await screen.findByRole('table');
 
+        expect(container.firstChild).toMatchSnapshot();
+    });
+
+    test('renders a table with correct number of rows', async () => {
+        render(<Products />);
+        // Wait for table to be rendered
+        const productsTable = await screen.findByRole('table');
         const rowsLength = screen.getAllByRole('row').length - 1; // Rows - 1 (table header)
 
-        expect(container.firstChild).toMatchSnapshot();
-
-        // Confirm table has been rendered with correct number of rows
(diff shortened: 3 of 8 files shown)


.github/workflows/validate-workflow.yml
```diff
@@ -22,6 +22,11 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v2
 
+    - name: Reconfigure git to use HTTP authentication
```
bc-chaz, line 25: was running into an issue with the `node-bigcommerce` package not being configured for ssh... this seems to be a workaround - https://github.com/actions/setup-node/issues/214#issuecomment-810829250.

Will also hitup Booker to see if he has any insight since he created the repo

README.md
```diff
@@ -8,7 +8,7 @@ This demo includes all of the files necessary to get started with a basic, hello
 
 To get the app running locally, follow these instructions:
 
-1. [Use Node 10+](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm#checking-your-version-of-npm-and-node-js)
+1. [Use Node 10+ and NPM 7+](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm#checking-your-version-of-npm-and-node-js)
```
chanceaclark, line 11: Should we add an `.nvmrc` file along with setting the [`engines`](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#engines) field in `package.json`?

package.json
```diff
@@ -15,15 +15,15 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
-    "@bigcommerce/big-design": "^0.27.0",
+    "@bigcommerce/big-design": "^0.30.2",
     "firebase": "^8.2.6",
     "jsonwebtoken": "^8.5.1",
     "mysql": "^2.18.1",
-    "next": "^10.2.3",
+    "next": "^11.1.1",
```
chanceaclark, line 22: Was there a reason for not bumping this to the latest? (`^12.0.4`)?

test/pages/products/index.spec.tsx
```diff
@@ -1,21 +1,26 @@
...
 
 describe('Product List', () => {
     test('renders correctly', async () => {
-        const { container } = render(<Products />);
+        const { container, getByRole } = render(<Products />);
         // Wait for table to be rendered
-        await screen.findByRole('table');
-
-        const rowsLength = screen.getAllByRole('row').length - 1; // Rows - 1 (table header)
+        await waitFor(() => getByRole('table'));
```
chanceaclark, line 11: We shouldn't be using `waitFor`. What we had before, using `screen` was better.

test/pages/products/index.spec.tsx
```diff
@@ -1,21 +1,26 @@
...
 
         expect(container.firstChild).toMatchSnapshot();
+    });
+
+    test('renders a table with correct number of rows', async () => {
+        const { getAllByRole, getByRole } = render(<Products />);
+        // Wait for table to be rendered
+        await waitFor(() => getByRole('table'));
```
chanceaclark, line 19: Same thing but also assign this to a const and use it in the `expect` below.
```suggestion
        const table = await screen.findByRole('table');
```

```tsx
// ...
expect(table).toBeDefined();
```

package.json
```diff
@@ -15,15 +15,15 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
-    "@bigcommerce/big-design": "^0.27.0",
+    "@bigcommerce/big-design": "^0.30.2",
```
chanceaclark, line 18: We should be including the other packages too instead of relying on hoisting here.

.github/workflows/validate-workflow.yml
chanceaclark: Curious, have we tried appending `github:` to the repo reference?
```json
{
  "node-bigcommerce": "github:bigcommerce/node-bigcommerce",
}
```

package.json
bc-chaz: yeah, 12+ has major breaking changes - to go down that route should be a separate PR - bumping to 11 doesn't include breaking changes and fixes the security vulnerability

test/pages/products/index.spec.tsx
bc-chaz: to be honest, I don't think it matters. `findBy`includes  `waitFor `. React testing library even includes this pattern in their docs (https://testing-library.com/docs/react-testing-library/example-intro#act). Outside of semantic preference, I don't believe there's any kind of performance hit one way or the other. The only real reason I can think of to use `screen` is to not list the queries as part of the render result, but that comes at the expense conciseness/ destructuring. 

The impetus behind this change was that `findByRole` was causing a `TypeError: Cannot read property 'createEvent' of null` error so I split it into `waitFor` and `getByRole`.

package.json
bc-chaz: the only other package I see is icons - I'm happy to include, though that's not necessarily within the purview of this PR (whose intent was to update to npm 7+). If we want to update/ modify / expand `package.json`, including a newer version of Next, that should be a separate PR.

test/pages/products/index.spec.tsx
chanceaclark: Kent C. Dodds writes about this https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-waitfor-to-wait-for-elements-that-can-be-queried-with-find. The main two points is that it's quicker to read (easier to debug) and you'll get a better error message using `findBy*` queries.

I believe there is an underlying issue for the `TypeError` I pulled your PR down and switched it over and wasn't receiving such error. (I was using node@16 + npm@8 though). We should dig into why you are getting that error.

package.json
chanceaclark: Yeah let's do it in another PR, should've added a 🍹 here.

I added this comment because when you install `@bigcommerce/big-design` it'll install both `@bigcommerce/big-design-icons` and `@bigcommerce/big-design-theme` as hoisted dependencies, since the core package depends on the two. I believe #70 works because it uses the hoisted dependency, but we should be explicitly defining `@bigcommerce/big-design-theme` in `dependencies` (I'll wait for this PR to add it).

chanceaclark (approved):

PR #73, opened 2021-11-19
feat(common): APPEX-202 implement orders page designs

bc-zachary:
## What?
Dependent on #71 
Updates `pages/orders/[orderId]` to match new designs. Includes new api routes for fetching shipping addresses and products, which are not included in the order data.

Screen Shot 2021-11-19 at 11 34 39 AM

## Why?
Jira: [APPEX-202](https://jira.bigcommerce.com/browse/APPEX-202)

## Testing / Proof
Tests passing and verified locally.


types/order.ts
```diff
@@ -27,4 +33,9 @@ export interface Order {
     shipping_cost_ex_tax: string;
     total_inc_tax: string;
     total_tax: string;
-}
\ No newline at end of file
+}
+
+export interface ShippingOrder {
```
chanceaclark, line 38: Let's extend this from `Order` since the and rename it to be more explicit. Here's my recommendation:
```ts
export interface ShippingAndProductsOrder extends Order {
  //...
}
```

lib/hooks.ts
```diff
@@ -67,11 +67,22 @@ export const useOrder = (orderId: number) => {
     const shouldFetch = context && orderId !== undefined;
 
     // Conditionally fetch product if it doesn't exist in the list (e.g. deep linking)
-    const { data, error } = useSWR<Order, ErrorProps>(shouldFetch ? [`/api/orders/${orderId}`, params] : null, fetcher);
+    const { data, error: orderError } = useSWR<Order, ErrorProps>(shouldFetch ? [`/api/orders/${orderId}`, params] : null, fetcher);
+    // Shipping addresses and products are not included in the order data and need to be fetched separately
+    const { data: shippingAddresses, error: shippingError } = useSWR(shouldFetch ? [`/api/orders/${orderId}/shipping_addresses`, params] : null, fetcher);
+    const { data: products, error: productsError } = useSWR(shouldFetch ? [`/api/orders/${orderId}/products`, params] : null, fetcher);
```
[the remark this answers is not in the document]
chanceaclark: Can we also split this into another hook to decouple the product fetching? This now means that a fast call to get orders now has to get this other information too.
```ts
const useShippingAndProductsOrder = (orderId: number) => {
  const { order } = useOrder(orderId);
  // ...
}
```

types/order.ts
```diff
@@ -27,4 +33,9 @@ export interface Order {
...
+    shipping_addresses: BillingAddress[];
```
chanceaclark, line 39: We should define a new type, even if they overlap `ShippingAddress !== BillingAddress`. The types may look and act similar, but they speak different languages.

types/order.ts
```diff
@@ -11,6 +11,12 @@ export interface BillingAddress {
     country: string;
 }
 
+export interface OrderProduct {
+    name: string;
```
chanceaclark, line 15: We should add `[key: string]: unknown;` to show that there are additional fields to this interface.

chanceaclark (commented):
Sorry for the amount of comments. Let me know if you want to chat offline to discuss any of it.

types/order.ts
bc-zachary: Would it make sense to have a base `Address` type to extend from?

chanceaclark: Let's keep them separate. It's alright to have some duplication.

pages/orders/[orderId]/index.tsx
```diff
@@ -1,129 +1,85 @@
...
+    )
 
-const InternalOrderPage: React.FC<{ order: Order }> = ({ order }) => {
-  const { billing_address } = order;
-  
+    const renderProducts = (products: OrderProduct[]) => (
+        <>
+            {products.map(product => <Text key={product.name}>{product.name}</Text>)}
+        </>
+    )
```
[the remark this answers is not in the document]
Comment (bc-zachary) [reply pages/orders/[orderId]/index.tsx:41]: The design actually has margins between products. I think it makes sense to distinguish them as a list of items rather than a unit like the addresses.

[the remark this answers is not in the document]
Comment (chanceaclark) [reply pages/orders/[orderId]/index.tsx:41]: Sounds good, might want to use a `ul` + `li` then. We have a helper on the theme for that.
```tsx
const StyledUl = styled.ul`
    ${({ theme }) => theme.helpers.listReset}

    color: ${({ theme }) => theme.colors.secondary70};
`
```

components/header.tsx
```diff
@@ -6,20 +6,36 @@ import InnerHeader from './innerHeader';
 export const TabIds = {
     HOME: 'home',
     PRODUCTS: 'products',
+    ORDERS: 'orders',
 };
 
 export const TabRoutes = {
     [TabIds.HOME]: '/',
     [TabIds.PRODUCTS]: '/products',
+    [TabIds.ORDERS]: '/orders',
```
bc-chaz, line 15: orders shouldn't be a tabbed route (since we don't want it to appear in the main tabbed navigation).

components/header.tsx
```diff
@@ -6,20 +6,36 @@ import InnerHeader from './innerHeader';
...
 };
 
+export const RouteTitles = {
+    [TabIds.HOME]: 'Home',
+    [TabIds.PRODUCTS]: 'Products',
+    [TabIds.ORDERS]: 'Orders',
+};
+
+const HeaderlessRoutes = [
+    '/orders/[orderId]',
```
bc-chaz, line 25: this will never be hit since you've defined it in `InnerRoutes` and that will be hit first (in the conditional inside of useEffect)

components/header.tsx
```diff
@@ -6,20 +6,36 @@ import InnerHeader from './innerHeader';
...
+    '/orders/[orderId]/labels',
+    '/orders/[orderId]/modal',
+]
```
bc-chaz, line 28: small nit: missing semicolon

components/innerHeader.tsx
```diff
@@ -2,20 +2,23 @@ import { Box, Button, H1, HR, Text } from '@bigcommerce/big-design';
...
 
 const InnerHeader = () => {
     const router = useRouter();
     const { pid } = router.query;
     const { list = [] } = useProductList();
     const { name } = list.find(item => item.id === Number(pid)) ?? {};
 
-    const handleBackClick = () => router.push(TabRoutes[TabIds.PRODUCTS]);
+    const routeHead = router.pathname.split('/')[1];
+    const title = RouteTitles[routeHead];
```
bc-chaz, line 14: Given that this is an `innerHeader`, and homepage will never be inner, we could probably drop `Home` from `RouteTitles`. Also, since we're not actually consuming `RouteTitles` inside of the main header page, we should probably move that const to this file.

```jsx
export const RouteTitles = {
    PRODUCTS: 'Products',
    ORDERS: 'Orders',
};
```

components/header.tsx
```diff
@@ -6,20 +6,36 @@ import InnerHeader from './innerHeader';
 export const TabIds = {
     HOME: 'home',
     PRODUCTS: 'products',
+    ORDERS: 'orders',
```
bc-chaz, line 9: we should prob remove this (since orders isn't a tab)

components/innerHeader.tsx
```diff
@@ -2,20 +2,23 @@ import { Box, Button, H1, HR, Text } from '@bigcommerce/big-design';
...
+
+    const handleBackClick = () => router.push(TabRoutes[routeHead]);
```
bc-chaz, line 16: since `TabRoutes` is just the route with a forward slash (and given orders isn't actually a tabbed route), we could probably just eliminate the use of `TabRoute` here.
```jsx
    const basePath = router.pathname.split('/')[1];
    const handleBackClick = () => router.push(`/${basePath}`);
```

bc-zachary: This actually brings up a question about where we want this button to direct to. When on `/orders/[orderId]` the text says `Orders` But there isn't currently an orders page to navigate to. If we do plan to include an orders page at some point, it would make sense to keep the label but maybe have it navigate to the home page for now. If we don't intend on adding an orders page, it might make more sense to rename it or disable it like the "Create shipment" button.

lib/hooks.ts
```diff
@@ -66,7 +66,7 @@ export const useOrder = (orderId: number) => {
     const params = new URLSearchParams({ context }).toString();
     const shouldFetch = context && orderId !== undefined;
 
-    // Conditionally fetch orderId is defined
+    // Conditionally fetch product if it doesn't exist in the list (e.g. deep linking)
```
chanceaclark, line 69: This is the `useOrder` hook, not sure `product` applies here.

lib/hooks.ts
```diff
@@ -75,3 +75,25 @@ export const useOrder = (orderId: number) => {
...
     };
 }
+
+export const useShippingAndProductsOrder = (orderId: number) => {
+    const { context } = useSession();
+    const params = new URLSearchParams({ context }).toString();
+    const shouldFetch = context && orderId !== undefined;
+
+    // Conditionally fetch product if it doesn't exist in the list (e.g. deep linking)
+    const { data, error: orderError } = useSWR<Order, ErrorProps>(shouldFetch ? [`/api/orders/${orderId}`, params] : null, fetcher);
```
chanceaclark, line 85: Why don't we be extensible and use `useOrder` here?
```suggestion
    const { order, error: orderError } = useOrder(orderId);
```

lib/hooks.ts
```diff
@@ -75,3 +75,25 @@ export const useOrder = (orderId: number) => {
...
+    const { data: shippingAddresses, error: shippingError } = useSWR<ShippingAddress[], ErrorProps>(
+        shouldFetch ? [`/api/orders/${orderId}/shipping_addresses`, params] : null, fetcher
+    );
+    const { data: products, error: productsError } = useSWR<OrderProduct[], ErrorProps>(
+        shouldFetch ? [`/api/orders/${orderId}/products`, params] : null, fetcher
+    );
+
+    return {
+        order: { ...data, shipping_addresses: shippingAddresses, products },
+        isLoading: !data && !orderError && !shippingError && !productsError,
```
chanceaclark, line 96: Should we also check for `!shippingAddresses` & `!products` here too?

pages/orders/[orderId]/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+const StyledAddress = styled.address`
+    color: ${({ theme }) => theme.colors.secondary70};
+    font-style: normal;
+    line-height: ${({ theme }) => theme.lineHeight.small};
+`;
+
+/*
+ * TODO check with Chance about StyledUl and theme
+ */
```
Comment (chanceaclark) [inline pages/orders/[orderId]/index.tsx:19]: Comment?

chanceaclark (commented):
Looking good, few more comments.

pages/orders/[orderId]/index.tsx
Comment (bc-zachary) [reply pages/orders/[orderId]/index.tsx:19]: whoops!

lib/hooks.ts
```diff
@@ -75,3 +75,24 @@ export const useOrder = (orderId: number) => {
...
+    const { context } = useSession();
+    const params = new URLSearchParams({ context }).toString();
+    const shouldFetch = context && orderId !== undefined;
+
+    const { order, error: orderError } = useOrder(orderId);
+    // Shipping addresses and products are not included in the order data and need to be fetched separately
+    const { data: shippingAddresses, error: shippingError } = useSWR<ShippingAddress[], ErrorProps>(
+        shouldFetch ? [`/api/orders/${orderId}/shipping_addresses`, params] : null, fetcher
+    );
+    const { data: products, error: productsError } = useSWR<OrderProduct[], ErrorProps>(
```
bc-chaz, line 89: imho, there's a lot going on with this hook. It seems like we're trying to cram all of the logic here, rather than having multiple, smaller functions that do just one thing. This can create issues with readability and maintainability. Moreover, if we're trying to package multiple API calls into one object, why not do that in the API level (via Promise.all)? That would seem like a more appropriate place to bundle the `shippingAddress` and `products` call, rather than creating multiple useSWR hooks here. Doing so in the API level will also allow us to simplify our `isLoading`, and `error` objects below.

lib/hooks.ts
```diff
@@ -75,3 +75,24 @@ export const useOrder = (orderId: number) => {
         error,
     };
 }
+
+export const useShippingAndProductsOrder = (orderId: number) => {
+    const { context } = useSession();
+    const params = new URLSearchParams({ context }).toString();
+    const shouldFetch = context && orderId !== undefined;
+
+    const { order, error: orderError } = useOrder(orderId);
```
bc-chaz, line 84: since `useOrder` is already a hook, why not just call directly where we need it in `pages`, rather than adding additional complexity here?

pages/orders/[orderId]/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+import { useRouter } from 'next/router';
+import React from 'react';
+import styled from 'styled-components';
+
+import ErrorMessage from '../../../components/error';
+import Loading from '../../../components/loading';
+import { useShippingAndProductsOrder } from '../../../lib/hooks';
+import { BillingAddress, OrderProduct, ShippingAndProductsOrder } from '../../../types';
+
+const StyledAddress = styled.address`
```
Comment (bc-chaz) [inline pages/orders/[orderId]/index.tsx:11]: I think wherever possible, we should prefer to use BigDesign's default functionality.  Here, this largely looks similar to BD's `Text` component. We could probably add `marginVertical` to get the effect you're after with `line-height` if need be.

pages/orders/[orderId]/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+                        },
+                    ]}
+                    items={items}
+                /> 
+            </Panel>
+        </>
+    );
+};
+
+const OrderPage: React.FC = () => {
```
Comment (bc-chaz) [inline pages/orders/[orderId]/index.tsx:80]: please remove `React.FC`, it's not necessary.
```suggestion
const OrderPage = () => {
```

pages/orders/[orderId]/index.tsx
```diff
@@ -0,0 +1,96 @@
+import { Button, Flex, H1, Panel, StatefulTable } from '@bigcommerce/big-design';
+import { useRouter } from 'next/router';
+import React from 'react';
```
Comment (bc-chaz) [inline pages/orders/[orderId]/index.tsx:3]: we don't need to import React for Next.js components

pages/orders/[orderId]/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+`;
+
+const StyledUl = styled.ul`
+    ${({ theme }) => theme.helpers.listReset}
+
+    color: ${({ theme }) => theme.colors.secondary70};
+    line-height: ${({ theme }) => theme.lineHeight.xxLarge};
+`;
+
+const InternalOrderPage = (order: ShippingAndProductsOrder) => {
```
Comment (bc-chaz) [inline pages/orders/[orderId]/index.tsx:24]: a part of me wonders if we should move this functional component to its own component page. Traditionally, with class based components, we'd have only one component per page. With hooks/ FC, that line has become a bit blurry insomuch that might only do it if the multiple components contain state and/ or side effects. 

What makes me think about it here is that this component itself is rendering multiple smaller components and is being called by `OrderPage` which itself has state that's managed by SWR. FWIW, I believe we're doing something similar in `products/[pid]`.

pages/orders/[orderId]/index.tsx
Comment (chanceaclark) [reply pages/orders/[orderId]/index.tsx:11]: I believe he is using the `<address>` element for accessibly semantics; at least that's what I would do. The rest of the styles is to clone the `Text` component.

pages/orders/[orderId]/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+};
+
+const OrderPage: React.FC = () => {
+    const router = useRouter();
+    const { orderId } = router.query;
+    const { isLoading, order, error }= useShippingAndProductsOrder(parseInt(`${orderId}`, 10));
+
+    if (isLoading) {
+        return <Loading />;
+    }
```
Comment (bc-chaz) [inline pages/orders/[orderId]/index.tsx:87]: 🍹 these multiple line returns on basic conditionals are driving me nuts... lol we're only doing this on the orders pages and imho isn't consistent with the rest of the app. I think ideally we should do one or the other and be consistent.

```suggestion
    if (isLoading) return <Loading />;
```

Comment (bc-zachary) [reply pages/orders/[orderId]/index.tsx:87]: I'll make sure to include this in the linting update.

pages/orders/[orderId]/index.tsx
Comment (bc-chaz) [reply pages/orders/[orderId]/index.tsx:11]: yeah, we're not really implementing accessibility at this point. We discussed potentially adding in an additional step or having "recipes" as Sophi would like to say. imho if someone was using a screen reader, as soon as it hits `Ship to` it would be abundantly clear what this section is. 

Let's follow-up with Booker and design about accessibility. In the meantime, this could probably be `Text`.

pages/orders/[orderId]/index.tsx
Comment (bc-chaz) [reply pages/orders/[orderId]/index.tsx:87]: thank you 🙏 😬

lib/hooks.ts
```diff
@@ -75,3 +75,20 @@ export const useOrder = (orderId: number) => {
         error,
     };
 }
+
+export const useShippingAndProductsOrder = (orderId: number) => {
```
chanceaclark, line 79: 🍹 Does this name make sense? Since it's **only** fetching Shipping Info and Products could we rename this to `useOrderShippingAndProductInfo` or something of that sorts?

pages/orders/[orderId]/index.tsx
Comment (chanceaclark) [reply pages/orders/[orderId]/index.tsx:9]: The idea here is that we are requiring that `order` be populated before we render the page. It's a better practice to ensure that you have data passed into your components rather than allowing `Type | undefined` being passed in. This helps keeps unnecessary checks outside of our components. We have been taking this approach at BigCommerce within our NextJS applications. Ask me offline for some examples of this.

It's not necessarily bad to have multiple components in page, especially if one is a simple one for data fetching and one for rending. If we wanted to move to a single component per file approach it's a lot of lift since every file/folder in `/pages` in a NextJS application is a route. We have solutions in place in our internal applications, but we enforce the patterns via custom linting rules and TypeScript module resolution.

pages/api/orders/[orderId]/shipping_products.ts
```diff
@@ -0,0 +1,36 @@
...
+        const bigcommerce = bigcommerceClient(accessToken, storeHash, 'v2');
+
+        switch (method) {
+            case 'GET': {
+                const shipping_addresses = bigcommerce.get(`/orders/${orderId}/shipping_addresses`);
+                const products = bigcommerce.get(`/orders/${orderId}/products`);
+
+                const data = await Promise.all([shipping_addresses, products])
+
+                res.status(200).json({ shipping_addresses: data[0], products: data[1] });
```
Comment (chanceaclark) [inline pages/api/orders/[orderId]/shipping_products.ts:21]: 🍹 I don't think we need `Promise.all` here. Couldn't we just await on the get requests? Since we wrap it in a try/catch is should resolve/reject in a similar manner.
```suggestion
                const shipping_addresses = await bigcommerce.get(`/orders/${orderId}/shipping_addresses`);
                const products = await bigcommerce.get(`/orders/${orderId}/products`);

                res.status(200).json({ shipping_addresses, products });
```

chanceaclark (approved):

PR #78, opened 2021-12-09
[WIP] feat(common) APPEX-213: update linting rules

bc-zachary:
## What?
- Updates linting to extend [bigcommerce/eslint-config](https://github.com/bigcommerce/eslint-config).
- Auto-fix resulting linting errors
- Manually fix remaining errors

## Why?
Standardize our linting rules for contributors.
## Testing / Proof
Verified by running the linter, ensuring app builds and runs.


package.json
```diff
@@ -42,12 +44,10 @@
...
     "babel-plugin-styled-components": "^2.0.2",
-    "eslint": "^7.21.0",
-    "eslint-import-resolver-typescript": "^2.4.0",
-    "eslint-plugin-import": "^2.22.1",
-    "eslint-plugin-react": "^7.22.0",
-    "eslint-plugin-react-hooks": "^4.2.0",
+    "eslint": "^7.32.0",
+    "eslint-config-next": "^12.0.7",
     "jest": "^26.6.3",
+    "prettier": "^2.5.1",
```
chanceaclark, line 50: We don't need to include this.

package.json
```diff
@@ -42,12 +44,10 @@
     "@typescript-eslint/parser": "^4.16.1",
     "babel-jest": "^26.6.3",
     "babel-plugin-styled-components": "^2.0.2",
-    "eslint": "^7.21.0",
-    "eslint-import-resolver-typescript": "^2.4.0",
-    "eslint-plugin-import": "^2.22.1",
-    "eslint-plugin-react": "^7.22.0",
-    "eslint-plugin-react-hooks": "^4.2.0",
+    "eslint": "^7.32.0",
```
chanceaclark, line 47: I believe we need this to be `^8.0.0`, so let's just bump to the latest.

package.json
```diff
@@ -42,12 +44,10 @@
     "@typescript-eslint/parser": "^4.16.1",
```
chanceaclark, line 44: I believe this is now a peer dependency which requires it to be `^5.0.0` (latest `^5.6.0`).

chanceaclark: Also, remove `@typescript-eslint/eslint-plugin` since it's not needed (included in `@bigcommerce/eslint-config`)

chanceaclark (commented):
Few comments...

I believe the reason the build is failing is due to it being on node v10. We might need to bump the CI job to v14 or something.

bc-zachary (commented):
Does it make sense to fix the build error in this first commit or roll it in with the manual fixes?

tsconfig.json
```diff
@@ -9,7 +9,7 @@
     "allowJs": true,
     "baseUrl": ".",
     "skipLibCheck": true,
-    "strict": false,
+    "strictNullChecks": true,
```
chanceaclark, line 12: Why did we change this? This change causes the build error. We shouldn't be touching the tsconfig in the PR as far as I'm aware.

As an aside, 🍹, we should be using `"strict": true` which enforces `strictNullChecks` along with some other ones. It ensures type safety throughout the codebase.

.github/workflows/validate-workflow.yml
```diff
@@ -16,7 +16,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [10.x]
+        node-version: [14.x]
```
chanceaclark, line 19: @bc-chaz and I were discussing this yesterday in standup. I believe the consensus was that if we want to enforce node v14, we'll have to go and update dev documentation alongside of it. We might have to hold on this change for now until we determine what we want to do.

.eslintrc.js
```diff
@@ -0,0 +1,5 @@
+require('@bigcommerce/eslint-config/patch');
```
chanceaclark, line 1: IIRC this is really for monorepos and our internal microapps using our ng-build. I don't think we need this here.

chanceaclark (commented):
@bc-zachary see the `strictNullChecks` comment about the build error.

.github/workflows/validate-workflow.yml
bc-chaz: unfortunately, if the build is failing, we might have to update this bit. I'm less concerned about updating the workflow - since this only affects the checks on GH, unlike nvmrc or potentially engines.

all that said, we should hold off on merging this PR till we talk to dev docs and figure out a good path forward. I'm going to add a `do not merge` label to this PR.

.eslintrc.js
bc-zachary: @chanceaclark Removing this require causes ESLint to throw `ESLint couldn't find the plugin "@typescript-eslint/eslint-plugin". (The package "@typescript-eslint/eslint-plugin" was not found when loaded as a Node module from the directory "~/.../sample-app-nodejs".)` then prompts to install `@typescript-eslint/eslint-plugin`. Seems like we need one or the other.

bc-chaz:
Closing for now - we can revisit later

PR #81, opened 2022-01-10
feat(common): APPEX-204 add api support for testing

bc-chaz:
## What?
Adds support for testing in different environments. Note: we cannot rely on `process.env.NODE_ENV` as `next start` will return 'production.'

## Why?
We need to be able to utilize testing api hosts on non-production targets.

## Testing / Proof
Verified by installing/ loading the app, linting, established production build and TypeScript by running npm run build.

### DIFF
--- a/lib/auth.ts
+++ b/lib/auth.ts
@@ -1,10 +1,17 @@
 import * as jwt from 'jsonwebtoken';
 import { NextApiRequest } from 'next';
 import * as BigCommerce from 'node-bigcommerce';
-import { QueryParams, SessionContextProps, SessionProps } from '../types';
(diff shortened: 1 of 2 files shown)


lib/auth.ts
```diff
@@ -1,6 +1,7 @@
 import * as jwt from 'jsonwebtoken';
 import { NextApiRequest } from 'next';
 import * as BigCommerce from 'node-bigcommerce';
+import { testApiConfig } from '../test/utils';
```
christensenep, line 4: I feel like calling this "test" implies that it's only used for automated tests, when actually this is just a general api config, isn't it? (I have to be honest I actually thought we'd already basically added this). Any reason not to just call it apiConfig?

christensenep (approved):

PR #93, opened 2022-04-13
Updated typescript to latest version

davido242:
## What?
A description about what this pull request implements and its purpose. Try to be detailed and describe any technical details to simplify the job of the reviewer and the individual on production support.

## Why?
...

## Testing / Proof
...

### DIFF
--- a/components/loading.tsx
+++ b/components/loading.tsx
@@ -2,7 +2,7 @@ import { Flex, H3, Panel, ProgressCircle } from '@bigcommerce/big-design';
 
 const Loading = () => (
     <Panel>
-        <H3>Hamari Loads....</H3>
+        <H3>Loads...</H3>
         <Flex justifyContent="center" alignItems="center">
             <ProgressCircle size="large" />
         </Flex>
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -47,7 +48,6 @@
     "eslint-plugin-import": "^2.22.1",
     "eslint-plugin-react": "^7.22.0",
     "eslint-plugin-react-hooks": "^4.2.0",
-    "jest": "^26.6.3",
-    "typescript": "^4.1.3"
```
bc-chaz, line 51: I believe TS as a dev dependency makes sense as it's precompile

bc-chaz, line 51: I believe TS as a dev dependency makes sense as it's precompile

components/loading.tsx
```diff
@@ -2,7 +2,7 @@ import { Flex, H3, Panel, ProgressCircle } from '@bigcommerce/big-design';
 
 const Loading = () => (
     <Panel>
-        <H3>Loading...</H3>
```
bc-chaz, line 5: since this is a loading indicator we should probably leave the wording as `loading`

bc-chaz:
@davido242 could you please add a brief description and then rebase to eliminate the conflicting files? other than that, we'd be happy to update TS

bc-chaz:
@davido242 could you please add a brief description and then rebase to eliminate the conflicting files? other than that, we'd be happy to update TS

bc-chaz:
Fixed in https://github.com/bigcommerce/sample-app-nodejs/pull/100

PR #95, opened 2022-04-28
fix(common): APPEX-279 delete all store users on uninstall

bc-zachary:
## What?
- Adds `deleteStoreUsers` method to both firebase and mysql modules to be called from `removeDataStore` on uninstall. - - Fixes `deleteStore` method.

## Why?
[APPEX-279](https://bigcommercecloud.atlassian.net/browse/APPEX-279)
...

## Testing / Proof
Verified manually for both Firebase and MySQL.
...

### DIFF
--- a/lib/auth.ts
+++ b/lib/auth.ts
@@ -84,10 +84,10 @@ export function decodePayload(encodedContext: string) {
     return jwt.verify(encodedContext, JWT_KEY);
 }
 
-// Removes store and storeUser on uninstall
+// Removes store and storeUsers on uninstall
 export async function removeDataStore(session: SessionProps) {
     await db.deleteStore(session);
-    await db.deleteUser(session);
+    await db.deleteStoreUsers(session);
 }
 
 // Removes users from app - getSession() for user will fail after user is removed
--- a/lib/dbs/firebase.ts
+++ b/lib/dbs/firebase.ts
@@ -1,5 +1,17 @@
 import { initializeApp } from 'firebase/app';
-import { deleteDoc, doc, getDoc, getFirestore, setDoc, updateDoc } from 'firebase/firestore';
+import {
+    collection,
+    deleteDoc,
+    doc,
+    getDoc,
+    getDocs,
+    getFirestore,
+    query,
+    setDoc,
+    updateDoc,
+    where,
+    writeBatch,
+} from 'firebase/firestore';
 import { SessionProps, UserData } from '../../types';
 
 // Firebase config and initialization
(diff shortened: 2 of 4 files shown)


chanceaclark (dismissed):

lib/auth.ts
```diff
@@ -87,7 +87,7 @@ export function decodePayload(encodedContext: string) {
 // Removes store and storeUser on uninstall
```
matthewvolk, line 87: 🍹 Not a big deal, but consider updating the comment to reflect

matthewvolk (approved):

matthewvolk (dismissed):

matthewvolk (approved):

lib/dbs/firebase.ts
```diff
@@ -91,6 +103,20 @@ export async function deleteUser({ context, user, sub }: SessionProps) {
     await deleteDoc(ref);
 }
 
+export async function deleteStoreUsers({ context, sub }: SessionProps) {
+    const contextString = context ?? sub;
+    const storeHash = contextString?.split('/')[1] || '';
+    const storeUsersRef = query(collection(db, 'storeUsers'), where('storeHash', '==', storeHash));
+    const storeUsers = await getDocs(storeUsersRef);
+    const batch = writeBatch(db);
```
bc-chaz, line 111: @bc-zachary @christensenep - we likely should not be removing collections like this. There's a couple of thoughts here. First and foremost, we always have to be careful when dealing with databases, especially with bulk transactions. Google actually goes as so far to explicitly say we should not be using this pattern with web apps as it has major security implications - they even exclude the Firebase v9 sample code (which would actually look somewhat similar to this) - https://firebase.google.com/docs/firestore/manage-data/delete-data#web.

Second, security implications aside, we likely should have some basic error handling. We might also want to consider some additional batch config/ handling (e.g. batch size). As it is right now, we're processing everything in one batch.  This might be fine for a really small app, but could become problematic on a larger app.

christensenep: I don't actually think we *are* deleting a collection here, are we?

bc-chaz: It depends. theoretically, there could be one store in which case it will delete the entire collection. I think, though, the larger issue is the sort of bulk removal from an unsecured front-end JS SDK. Specifically, since this action is being taken from an accessible address, e.g. `storeHash.mybigcommerce.com/manage/app/APPID/api/uninstall`, we currently have no way of authenticating that the action is being taken from an authorized user, save for the JWT, but that's the same exact JWT that gets sent on `load` - meaning anyone who can load the app, or anyone with the JWT, could theoretically remove all of the users.

christensenep: Ok, so to sum up:

We have indeed found a vulnerabilty, not with the sample app specifically, but with our app callbacks generally, in which the jwt can be used interchangably with any endpoint (load/uninstall/install). Most notably this could be used to do weird stuff with the uninstall endpoints, but in theory accessing any of them with the wrong jwt is bad (and this would apply to any future endpoints we might add). This is captured [here](https://bigcommercecloud.atlassian.net/browse/APPEX-289)

Regarding the actual implementation of the uninstall callback here, though, are we on the same page that this implementation is fine? One thing I do think we may need to add is handling batches of over 500 users (I believe batches of 500 operations is the max for firebase). But I don't think there's anything fundamentally wrong/insecure about performing these deletions here. Notably, this is actually executed on the backend, not the frontend. Any reason not to revert the revert (possibly adding the 500 count case, although I don't personally think we need to wait on that).

bc-chaz: The only caveat here is that we will need to update the code to check that the JWT was called with uninstall - otherwise, we're merging in the original vulnerability. The 500 batch fix would be nice

christensenep: Oh that latter part I don't think we can do yet, need a change on the App Registry side of things (in the ticket above), the JWT simply lacks any information that says which endpoint its meant to be used with. I think we can probably tackle that once the above ticket is done.

PR #103, opened 2022-06-27
fix(common): corrects error in workflow

bc-chaz:
## What?
Fixes an error in the GH workflow.

## Why?
failing workflows

## Testing / Proof
GH


matthewvolk (approved):
LGTM 👍

PR #104, opened 2022-06-28
feat(common): upgrade next, react, big design, and supporting libraries

bc-chaz:
## What?
Upgrades NextJS to 12. As part of the update, this gives a minor bump to React (we're unable to upgrade React to 18 as Big-Design does not yet support React v18). This PR also updates Big Design theme, as well as most of the dev dependencies (testing, eslint, etc).

## Why?
Security vulnerabilities

## Testing / Proof
verified locally and by running the app on a prod store; also ran linting, unit tests, and confirmed build passing.


lib/db.ts
```diff
@@ -1,18 +1,20 @@
...
+import * as firebaseDB from './dbs/firebase';
+import * as sqlDB from './dbs/mysql';
 
 const { DB_TYPE } = process.env;
 
 let db: Db;
 
 switch (DB_TYPE) {
     case 'firebase':
-        db = require('./dbs/firebase');
```
bc-chaz, line 9: this seemed to be creating issues with webpack5. was receiving an error, `WEBPACK_IMPORTED_MODULE_2__.default.setUser is not a function`

components/header.tsx
```diff
@@ -58,8 +58,8 @@ const Header = () => {
     });
 
     const items = [
-        { id: TabIds.HOME, title: 'Home' },
-        { id: TabIds.PRODUCTS, title: 'Products' },
+        { ariaControls: 'home', id: TabIds.HOME, title: 'Home' },
```
bc-chaz, line 61: curiously, BigDesign lists this field as optional but is throwing a warning from https://github.com/bigcommerce/big-design/blob/master/packages/big-design/src/components/Tabs/Tabs.tsx#L40. I'm adding this in for now, but doesn't seem to fit our particular use case of using tabs for navigation.

types/db.ts
```diff
@@ -12,12 +12,11 @@ export interface UserData {
 }
 
 export interface Db {
-    hasStoreUser(storeHash: string, userId: string): boolean;
+    hasStoreUser(storeHash: string, userId: string): Promise<boolean> | boolean;
     setUser(session: SessionProps): Promise<void>;
     setStore(session: SessionProps): Promise<void>;
     setStoreUser(session: SessionProps): Promise<void>;
-    getStore(): StoreData | null;
```
bc-chaz, line 19: `getStore` no longer exists - not sure how this was passing previously

package.json
```diff
@@ -15,39 +15,40 @@
   "author": "",
   "license": "ISC",
   "engines": {
-    "node": ">=12.1.0 <19",
+    "node": ">=12.22.0 <19",
     "npm": ">=7.20.x <8.14"
   },
   "dependencies": {
     "@bigcommerce/big-design": "^0.30.2",
```
bc-chaz, line 22: capturing upgrading BigDesign in a separate ticket/ PR

package.json
```diff
@@ -15,39 +15,40 @@
   "author": "",
   "license": "ISC",
   "engines": {
-    "node": ">=12.1.0 <19",
+    "node": ">=12.22.0 <19",
```
matthewvolk, line 18: Out of curiosity, is there any reason we cap Node versions at 18.9? I see the `>=12.22.0` is from Next.js v12, but is there anything in Node 19 that would break this app?

matthewvolk (approved):
LGTM 👍 Tested the README.md flow on the new branch locally and everything is still working perfectly.

package.json
bc-chaz: node v19 hasn't been released yet so it is untested. Once v19 is released, we can test and update the engines

PR #105, opened 2022-06-30
feat(common): appex-274 step 4 upgrade next, react, firebase and other libraries

bc-chaz:
## What?
STEP 4: Upgrades NextJS to 12. As part of the update, this gives a minor bump to React (we're unable to upgrade React to 18 as Big-Design does not yet support React v18). This PR also updates Firebase, styled-components, as well as most other dependencies.

## Why?
Security vulnerabilities

## Testing / Proof


matthewvolk (approved):

bc-zachary (approved):
👍

PR #106, opened 2022-06-30
feat(common) appex-274 step 3 upgrade next, react, firebase, and other libraries

bc-chaz:
## What?
STEP 3: Upgrades NextJS to 12. As part of the update, this gives a minor bump to React (we're unable to upgrade React to 18 as Big-Design does not yet support React v18). This PR also updates Firebase, styled-components, as well as most other dependencies.

## Why?
Security vulnerabilities

## Testing / Proof

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ This demo includes all of the files necessary to get started with a basic, hello
 
 To get the app running locally, follow these instructions:
 
-1. [Use Node 10+](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm#checking-your-version-of-npm-and-node-js)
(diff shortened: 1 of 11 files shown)


matthewvolk (approved):
I think you might have already mentioned it/are aware, but certain changes included in this  PR will make some of the code "out of sync" with the tutorial ([example](https://developer.bigcommerce.com/docs/ZG9jOjE4MjIyODMz-step-3-integrate-the-big-commerce-api-and-add-a-database#configure-the-firebase-database)).

Outside of that, this LGTM!

bc-zachary (approved):
lgtm 👍

PR #107, opened 2022-06-30
feat(common) appex-274 step 2 upgrade next, react, and BigDesign

bc-chaz:
## What?
STEP 2: Upgrades NextJS to 12. As part of the update, this gives a minor bump to React (we're unable to upgrade React to 18 as Big-Design does not yet support React v18). This PR also updates BigDesign, styled-components, as well as most other dependencies.

## Why?
Security vulnerabilities

## Testing / Proof


matthewvolk (changes requested):
I think this PR might just be missing the quick update to use Node 12+ in the README.md. After that, this one looks good as well!

matthewvolk (approved):

PR #127, opened 2022-09-16
Automated billing beta

bc-chaz:
## What?
Automated billing beta

## Why?
demo

## Testing / Proof
locally, confirmed unit tests, linting, and build passing


lib/checkout.ts
```diff
@@ -0,0 +1,171 @@
...
+                    product: {
+                        entityId: `${checkoutAppId}`,
+                        type: 'APPLICATION'
+                    },
+                    scope: {
+                        entityId: `${storeHash}`,
+                        type: 'STORE'
+                    },
+                    pricingPlan: planItems[productId].pricingPlan,
+                    redirectUrl: `https://store-${storeHash}.my-staging.zone/manage/app/34030/${appId}/${productId}`,
```
christensenep, line 116: Should we make this an env variable so it can work in arbitrary environments?

christensenep (approved):

lib/checkout.ts
bc-chaz: ahhh might be tricky if only because `.env` doesn't know the storeHash (unless we manually add it). Conversely, we can add a new variable that could capture the environment (e.g. staging)... I might go this route as we can then reuse the environment variable for api and login vars

PR #129, opened 2022-09-23
feat(common): adds sql support

bc-chaz:
## What?
Adds mysql support to the checkout reference application.

## Why?
Needed for release

## Testing / Proof
deployed to Heroku, verified clearDB populated with correct checkout and plan info. Also ran linting, tests, and confirmed build.

### DIFF
--- a/lib/dbs/mysql.ts
+++ b/lib/dbs/mysql.ts
@@ -1,6 +1,7 @@
 import * as mysql from 'mysql';
 import { promisify } from 'util';
 import { SessionProps, StoreData } from '../../types';
+import { trialDays } from '../checkout';
 
 const MYSQL_CONFIG = {
     host: process.env.MYSQL_HOST,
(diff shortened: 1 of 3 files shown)


lib/dbs/mysql.ts
```diff
@@ -91,3 +97,47 @@ export async function getStoreToken(storeHash: string) {
...
+
+// CHECKOUT Functions
+export async function setStorePlan(session: SessionProps) {
+    const { access_token: accessToken, context, plan, sub } = session;
+    // Only set on app install or subscription verification (load)
+    if ((!accessToken && !plan?.pid) || (plan && !plan.isPaidApp)) return null;
+
+    const contextString = context ?? sub;
+    const storeHash = contextString?.split('/')[1] || '';
+    const defaultEnd = Date.now() + (trialDays * 24 * 60 * 60 * 1000);
```
bc-chaz, line 109: this is exactly what we're doing for Firebase... should we be using sql's timestamp here and then convert later? or is this fine for now? @christensenep

scripts/db.js
```diff
@@ -42,6 +43,27 @@ const storeUsersCreate = query('CREATE TABLE `storeUsers` (\n' +
...
 );
 
-Promise.all([usersCreate, storesCreate]).then(() => {
+const planCreate = query('CREATE TABLE `plan` (\n' +
+    '  `id` int(11) unsigned NOT NULL AUTO_INCREMENT,\n' +
+    '  `storeHash` varchar(10) NOT NULL,\n' +
+    '  `pid` varchar(20) NOT NULL,\n' +
+    '  `isPaidApp` boolean,\n' +
+    '  `showPaidWelcome` boolean,\n' +
+    '  `trialEndDate` bigint(20),\n' +
```
bookernath, line 52: datetime type would make more sense to me

bookernath (approved):
LGTM, next reasonable iteration would be to introduce an ORM so we don't have to mess with raw SQL so much

PR #130, opened 2022-09-27
Appex 406

bc-chaz:
## What?
Updates reference implementation to dynamically fetch merchant account uuid, as opposed to hardcoding it.

Note: this PR includes sql changes until https://github.com/bigcommerce/sample-app-nodejs/pull/129 is merged.

## Why?
Beta

## Testing / Proof
verified locally, ran lint, tests, and build

### DIFF
--- a/pages/api/checkout/[pid].ts
+++ b/pages/api/checkout/[pid].ts
@@ -1,13 +1,16 @@
 import { NextApiRequest, NextApiResponse } from 'next';
-import { getSession, setCheckout } from '../../../lib/auth';
+import { bigcommerceClient, getSession, setCheckout } from '../../../lib/auth';
 import { getCheckoutBody } from '../../../lib/checkout';
 
 export default async function checkout(req: NextApiRequest, res: NextApiResponse) {
     const { query: { pid } } = req;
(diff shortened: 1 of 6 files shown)


pages/api/checkout/[pid].ts
```diff
@@ -1,13 +1,16 @@
...
 
 export default async function checkout(req: NextApiRequest, res: NextApiResponse) {
     const { query: { pid } } = req;
 
     try {
-        const { storeHash } = await getSession(req);
-        const checkoutBody = getCheckoutBody(String(pid), storeHash);
+        const { accessToken, storeHash } = await getSession(req);
+        const bigcommerce = bigcommerceClient(accessToken, storeHash, 'v2');
+        const { account_uuid: merchantUuid } = await bigcommerce.get(`/store`);
```
Comment (bookernath) [inline pages/api/checkout/[pid].ts:11]: 🍹 what happens if this call fails?

Comment (bc-chaz) [reply pages/api/checkout/[pid].ts:11]: should be caught in the catch below - we definitely need better error handling though. We have a ticket for that.

bookernath (approved):

PR #133, opened 2022-10-27
feat(common): appex-447 updates big-design/firebase/styled-components

bc-zachary:
## What?
Updates big-design, firebase and styled-components dependencies and updates snapshots for BD and styled-components.

## Why?
Snyk alerts

## Testing / Proof
Tests passing and verified running/build/lint locally

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "dev": "next",
     "build": "next build",
     "start": "next start -p $PORT",
-    "test": "jest",
+    "test": "jest --updateSnapshot",
     "lint": "eslint . --ext .ts,.tsx,.js",
     "db:setup": "node scripts/db.js"
   },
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -38,7 +38,7 @@
     "@types/jest": "^27.5.2",
     "@types/node": "^18.0.0",
     "@types/react": "^17.0.2",
-    "@types/styled-components": "^5.1.25",
+    "@types/styled-components": "^5.1.26",
```
bc-chaz, line 41: while we're here, should we consider upgrading to the latest version? i.e. I believe we're on 5.3.6 - from what I can tell there aren't any breaking changes

package.json
```diff
@@ -19,16 +19,16 @@
     "npm": ">=7.20.x <8.14"
   },
   "dependencies": {
-    "@bigcommerce/big-design": "^0.32.1",
+    "@bigcommerce/big-design": "^0.33.0",
```
bc-chaz, line 22: qq: does this break anything if we use the latest version of BD? `0.34.2`

package.json
```diff
@@ -19,16 +19,16 @@
...
     "@bigcommerce/big-design-theme": "^0.16.0",
-    "firebase": "^9.9.2",
+    "firebase": "^9.9.3",
```
bc-chaz, line 24: we might want to give this a slight bump as well - there's a `9.9.4` that's probably pretty safe

package.json
bc-zachary: Nope! I'll push that up. 👍

package.json
bc-zachary: I think with the carat `^` preceding the version, `npm i` automatically downloads the latest minor update — which in this case is `9.13.0` — and it's building fine for me locally so I think I could even bump it to that unless we want to be able to use `9.9.x`

package.json
```diff
@@ -19,16 +19,16 @@
...
     "jsonwebtoken": "^8.5.1",
     "mysql": "^2.18.1",
     "next": "^12.2.5",
```
matthewvolk, line 27: Maybe we should make a ticket to bump us to Next 13! Next.js Conf earlier this week has me like: 🤤😍

matthewvolk (commented):
Awesome! Will approve as soon as those minor bumps mentioned above are pushed.

matthewvolk (approved):
LGTM 👍

PR #145, opened 2022-12-22
feat(appext): APPEX-359 Add refernce example using App Extensions

bc-nekisham:
## What?
Adding in the ability to see an example of App Extensions with the already loaded products list on headerless route product/productId

## Why?
To give app developers  a reference  of App Extensions implementation of it in the Node Sample App

## Testing / Proof
...

https://user-images.githubusercontent.com/109627308/209204736-d68be80b-2ccb-4279-8c38-578633f21c1c.mov

### DIFF
--- a/components/header.tsx
+++ b/components/header.tsx
@@ -17,6 +17,7 @@ const HeaderlessRoutes = [
     '/orders/[orderId]',
     '/orders/[orderId]/labels',
     '/orders/[orderId]/modal',
+    '/product/[productId]',
 ];
 
 const InnerRoutes = [
--- a/pages/product/[productId]/index.tsx
+++ b/pages/product/[productId]/index.tsx
@@ -1,0 +1,47 @@
+import { Panel, Text } from "@bigcommerce/big-design";
+import { useRouter } from "next/router";
+import ErrorMessage from "@components/error";
+import Loading from "@components/loading";
+import { useProductInfo, useProductList } from "@lib/hooks";
+
+const SidePanel = () => {
+  const router = useRouter();
+  const productId = Number(router.query?.productId);
+  const { error, isLoading, list = [] } = useProductList();
+  const { isLoading: isInfoLoading, product } = useProductInfo(productId, list);
+  const {
+    description,
+    is_visible: isVisible,
+    name,
+    price,
+    type,
+  } = product ?? {};
+  const panelData = { description, isVisible, name, price, type };
+
+  const capitalizeFirst = (str: string) => {
+    return str.charAt(0).toUpperCase() + str.slice(1);
+  };
+
+  if (isLoading || isInfoLoading) return <Loading />;
+  if (error) return <ErrorMessage error={error} />;
+
+  return (
+    <>
+      <Panel header="Basic Information">
+        <Text bold>Product name</Text>
+        <p>{panelData.name}</p>
+        <Text bold>Product type</Text>
+        <p>{capitalizeFirst(panelData.type)}</p>
+        <Text bold>Default price (excluding tax)</Text>
+        <p>$ {panelData.price}</p>
+        <p>{panelData.isVisible}</p>
+      </Panel>
+      <Panel header="Description">
+        <Text bold>Description</Text>
+        <p>{panelData.description}</p>
+      </Panel>
+    </>
+  );
+};
+
+export default SidePanel;
--- a/test/pages/__snapshots__/index.spec.tsx.snap
+++ b/test/pages/__snapshots__/index.spec.tsx.snap
@@ -2,59 +2,59 @@
 
 exports[`Homepage renders correctly 1`] = `
 <div
-  class="styled__StyledBox-sc-sj5f1m-0 cmToHM styled__StyledPanel-sc-1h6ef3q-0 dJwQJf"
+  class="styled__StyledBox-sc-sj5f1m-0 odkmb styled__StyledPanel-sc-1h6ef3q-0 gyZFEm"
   id="home"
 >
   <div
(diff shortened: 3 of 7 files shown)


components/header.tsx
```diff
@@ -17,6 +17,7 @@ const HeaderlessRoutes = [
     '/orders/[orderId]',
     '/orders/[orderId]/labels',
     '/orders/[orderId]/modal',
+    '/product/[productId]',
```
bc-zachary, line 20: 🍹 Maybe this has already been discussed, but this route name is kind of ambiguous. Could we come up with something a little more descriptive to differentiate this route from `/products/[pid]`?

pages/product/[productId]/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { useRouter } from "next/router";
+import ErrorMessage from "@components/error";
+import Loading from "@components/loading";
+import { useProductInfo, useProductList } from "@lib/hooks";
+
+const SidePanel = () => {
+  const router = useRouter();
+  const productId = Number(router.query?.productId);
+  const { error, isLoading, list = [] } = useProductList();
+  const { isLoading: isInfoLoading, product } = useProductInfo(productId, list);
```
Comment (bc-zachary) [inline pages/product/[productId]/index.tsx:11]: 🍹 Seems like we could make `list` an optional parameter for `useProductInfo` or even just pass in an empty array so that we just fetch the one product instead of the whole list and then filtering it. Then we can get rid of `useProductList` and only have to check one `isLoading` condition. Might be for another ticket, just noticed this pattern and making a note.

bc-zachary (approved):
A couple non-blocking comments but LGTM otherwise 👍

components/header.tsx
bc-nekisham: Oh yeah I spoke with Chaz about the route name and we went this one, the change from pid to productId was to differentiate it, front the other route.

pages/product/[productId]/index.tsx
Comment (bc-nekisham) [reply pages/product/[productId]/index.tsx:11]: Ohh yeah that could be a could idea, will keep this in mind and bring it up for a future change to improve this addition!

pages/product/[productId]/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const {
+    description,
+    is_visible: isVisible,
+    name,
+    price,
+    type,
+  } = product ?? {};
+  const panelData = { description, isVisible, name, price, type };
```
Comment (bc-chaz) [inline pages/product/[productId]/index.tsx:19]: this seems odd that we're taking destructured elements only to place them back into an object - we're effectively reproducing `product` here...

pages/product/[productId]/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+  };
+
+  if (isLoading || isInfoLoading) return <Loading />;
+  if (error) return <ErrorMessage error={error} />;
+
+  return (
+    <>
+      <Panel header="Basic Information">
+        <Text bold>Product name</Text>
+        <p>{panelData.name}</p>
```
Comment (bc-chaz) [inline pages/product/[productId]/index.tsx:32]: instead of `p` element, we should prefer to use BigDesign `Text`

pages/product/[productId]/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+        <Text bold>Product type</Text>
+        <p>{capitalizeFirst(panelData.type)}</p>
+        <Text bold>Default price (excluding tax)</Text>
+        <p>$ {panelData.price}</p>
```
Comment (bc-chaz) [inline pages/product/[productId]/index.tsx:36]: usually it's a bit of a smell that we need to destructure an object if we have several references to object.prop

pages/product/[productId]/index.tsx
```diff
@@ -0,0 +1,47 @@
+import { Panel, Text } from "@bigcommerce/big-design";
+import { useRouter } from "next/router";
+import ErrorMessage from "@components/error";
+import Loading from "@components/loading";
+import { useProductInfo, useProductList } from "@lib/hooks";
+
+const SidePanel = () => {
```
Comment (bc-chaz) [inline pages/product/[productId]/index.tsx:7]: `SidePanel` is a bit ambiguous - we might want to call this `ProductAppExtension`

components/header.tsx
bc-chaz: yeah, the more I think about this, the more I think maybe we should change this a bit to be more descriptive/ clear - maybe `productAppExtension`

pages/product/[productId]/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+    is_visible: isVisible,
+    name,
+    price,
+    type,
+  } = product ?? {};
+  const panelData = { description, isVisible, name, price, type };
+
+  const capitalizeFirst = (str: string) => {
+    return str.charAt(0).toUpperCase() + str.slice(1);
+  };
```
Comment (bc-chaz) [inline pages/product/[productId]/index.tsx:23]: generally, I like having helper functions, though if we're only using a single line function in one place, it makes me wonder if it's necessary... we might just assign this to a variable

```suggestion
    const productType = type.replace(/^\w/, c => c.toUpperCase());
```

pages/product/[productId]/index.tsx
```diff
@@ -0,0 +1,47 @@
+import { Panel, Text } from "@bigcommerce/big-design";
+import { useRouter } from "next/router";
+import ErrorMessage from "@components/error";
+import Loading from "@components/loading";
+import { useProductInfo, useProductList } from "@lib/hooks";
+
+const SidePanel = () => {
+  const router = useRouter();
```
Comment (bc-chaz) [inline pages/product/[productId]/index.tsx:8]: @bc-nekisham note: we're using 4 space intending in this repo

bc-nekisham:
Updated changes here: https://github.com/bigcommerce/sample-app-nodejs/pull/148

PR #148, opened 2023-01-03
feat(appext): APPEX-461 Add reference example using App Extensions

bc-nekisham:
## What?
Adding in the ability to see an example of App Extensions with the already loaded products list on headerless route product/[productAppExtension]

## Why?
Adding in requested changes requested from (https://github.com/bigcommerce/sample-app-nodejs/pull/145)

## Testing / Proof
...

https://user-images.githubusercontent.com/109627308/210449769-2d48f7e9-c106-4645-9d0c-157cc48c4735.mov


components/header.tsx
```diff
@@ -17,7 +17,7 @@ const HeaderlessRoutes = [
     '/orders/[orderId]',
     '/orders/[orderId]/labels',
     '/orders/[orderId]/modal',
-    '/product/[productId]',
+    '/product/[productAppExtension]',
```
bc-chaz, line 20: ahh, I think `productId` here is correct - we want to change the name of the repo, not the query param

```suggestion
    '/productAppExtension/[productId]',
```

pages/product/[productAppExtension]/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+import Loading from "@components/loading";
+import { useProductInfo, useProductList } from "@lib/hooks";
+
+const ProductAppExtension = () => {
+    const router = useRouter();
+    const productAppExtension = Number(router.query?.productAppExtension);
+    const { error, isLoading, list = [] } = useProductList();
+    const { isLoading: isInfoLoading, product } = useProductInfo(productAppExtension, list);
+    const { description, is_visible: isVisible,name, price, type,} = product ?? {};
+
```
Comment (bc-chaz) [inline pages/product/[productAppExtension]/index.tsx:13]: 🍹 small nit: no extra line needed here

components/header.tsx
bc-nekisham: Ohhh okayyyy! Changing this back.

pages/product/[productAppExtension]/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+            <Text bold>Default price (excluding tax)</Text>
+            <Text>${price}</Text>
+            <Text>{isVisible}</Text>
+        </Panel>
+        <Panel header="Description">
+            <Text bold>Description</Text>
+            <Text>{description}</Text>
+        </Panel>
+    </>
+  );
```
Comment (bc-chaz) [inline pages/product/[productAppExtension]/index.tsx:35]: the spacing here is still wonky

```suggestion
    return (
        <>
            <Panel header="Basic Information">
                <Text bold>Product name</Text>
                <Text>{name}</Text>
                <Text bold>Product type</Text>
                <Text>{productType}</Text>
                <Text bold>Default price (excluding tax)</Text>
                <Text>${price}</Text>
                <Text>{isVisible}</Text>
            </Panel>
            <Panel header="Description">
                <Text bold>Description</Text>
                <Text>{description}</Text>
            </Panel>
        </>
    );
```

pages/product/[productAppExtension]/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+    
+    const productType = type?.replace(/^\w/, (c: string) => c.toUpperCase());
+
+    if (isLoading || isInfoLoading) return <Loading />;
+    if (error) return <ErrorMessage error={error} />;
+
+  return (
+        <>
+        <Panel header="Basic Information">
+            <Text bold>Product name</Text>
```
Comment (bc-chaz) [inline pages/product/[productAppExtension]/index.tsx:22]: rather than using `Text bold`, we might actually make use of header here to denote that this is actually a header for the text below. Please be sure to import `H4` from BigDesign

```suggestion
import { H4, Panel, Text } from "@bigcommerce/big-design";
...
            <H4>Product name</H4>
```

bc-chaz:
@bc-nekisham looking at the screenshot, it appears that the panel is rendering oddly - can we look into what's going on?

pages/product/[productAppExtension]/index.tsx
Comment (bc-nekisham) [reply pages/product/[productAppExtension]/index.tsx]: updated

bc-nekisham:
> @bc-nekisham looking at the screenshot, it appears that the panel is rendering oddly - can we look into what's going on?
the description panel was flowing over the app ext panel and the two panels were flowing into each other themselves- I changed the margins on both

https://user-images.githubusercontent.com/109627308/210641266-97764c0e-e99f-4719-8959-d8b5685733da.mov

test/pages/product/index.spec.tsx
```diff
@@ -1,18 +1,18 @@
...
-describe('SidePanel', () => {
+describe('ProductAppExtension', () => {
     const router = { query: { pid: '1' } };
     useRouter.mockReturnValue(router);
 
     test('renders correctly', async () => {
-        const { container } = render(<SidePanel />);
+        const { container } = render(<ProductAppExtension />);
         // Wait for table to be rendered
         screen.getByText('Basic Information');
```
bc-chaz, line 17: it seems odd that we have this test, but we're not actually awaiting anything

test/pages/product/index.spec.tsx
```diff
@@ -1,18 +1,18 @@
...
 jest.mock('next/router', () => ({
     useRouter: jest.fn(),
 }));
 
-describe('SidePanel', () => {
+describe('ProductAppExtension', () => {
     const router = { query: { pid: '1' } };
     useRouter.mockReturnValue(router);
 
     test('renders correctly', async () => {
```
bc-chaz, line 14: we have an opening `async`, but no `await` in this test

test/pages/product/index.spec.tsx
bc-nekisham: updated

test/pages/product/index.spec.tsx
bc-nekisham: updated

pages/productAppExtension/[productId]/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+import Loading from "@components/loading";
+import { useProductInfo, useProductList } from "@lib/hooks";
+
+const ProductAppExtension = () => {
+    const router = useRouter();
+    const productId = Number(router.query?.productId);
+    const { error, isLoading, list = [] } = useProductList();
+    const { isLoading: isInfoLoading, product } = useProductInfo(productId, list);
+    const { description, is_visible: isVisible,name, price, type,} = product ?? {};   
+    const productType = type?.replace(/^\w/, (c: string) => c.toUpperCase());
```
Comment (matthewvolk) [inline pages/productAppExtension/[productId]/index.tsx:13]: 🍹 I'd recommend renaming `productType` to be more descriptive of what this variable actually holds, for readability purposes. Consider naming it `typeCapitalized`, for example.

test/pages/productAppExtension/index.spec.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+    test('renders correctly', () => {
+        const { container } = render(<ProductAppExtension />);
+        const panelOneHeader = screen.getByText("Basic Information");
+
+        expect(container.firstChild).toMatchSnapshot();
+        expect(panelOneHeader).toBeInTheDocument();
+    });
+});
+
```
matthewvolk, line 22: 🍹 I might suggest testing for error states. For example, what happens if `router.query.pid` is undefined? What should the UI render in that case?

pages/productAppExtension/[productId]/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+            </Panel>
+            <Panel header="Description" margin="xxSmall">
+                <H4 margin="none">Description</H4>
+                <Text>{description}</Text>
+            </Panel>
+        </>
+    );
+};
+
+export default ProductAppExtension;
```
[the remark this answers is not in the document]
Comment (bc-nekisham) [reply pages/productAppExtension/[productId]/index.tsx]: ahh, thanks!

matthewvolk:
🍹 Not a big deal, but if you have a chance to modify the commit message to fix the typo and force push that up, that would be good.

pages/productAppExtension/[productId]/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+    return (
+        <>
+            <Panel header="Basic Information" margin="xxSmall">
+                <H4 margin="none">Product name</H4>
+                <Text>{name}</Text>
+                <H4 margin="none">Product type</H4>
+                <Text>{typeCapitalized}</Text>
+                <H4 margin="none">Default price (excluding tax)</H4>
+                <Text>${price}</Text>
+                <Text>{isVisible}</Text>
```
Comment (bc-chaz) [inline pages/productAppExtension/[productId]/index.tsx:27]: `isVisible` doesn't seem to be showing - let's also add a header for this

```suggestion
                <H4 margin="none">Visible on storefront</H4>
                <Text>{isVisible}</Text>
```

bc-nekisham:
Updated visible on storefront 

https://user-images.githubusercontent.com/109627308/211098871-acbd4d9e-2f2e-4e35-b03e-f104737fb76c.mov

lib/hooks.ts
```diff
@@ -61,6 +63,7 @@ export function useProductInfo(pid: number, list: ListItem[]) {
     };
 }
 
+
```
bc-chaz, line 66: nit: extra line

lib/hooks.ts
```diff
@@ -47,10 +47,12 @@ export function useProductList(query?: QueryParams) {
     };
 }
 
-export function useProductInfo(pid: number, list: ListItem[]) {
+export function useProductInfo(pid: number, list?:ListItem[]) {
     const { context } = useSession();
     const params = new URLSearchParams({ context }).toString();
-    const product = list.find(item => item.id === pid);
+
+    let product: any; if(list.length > 0) { list = product }
```
bc-chaz, line 54: let's move the conditional to its own line. There should only be one variable/ statement per line. We should also avoid using `any` if we can. The inside of the brackets is also incorrect... we should copy the assignment we deleted on ln53 here...

 ```suggestion
    let product: ListItem[]; 
    
    if (list.length) { 
        product = list.find(item => item.id === pid);
    }
```

File: lib/hooks.ts [hunk 13]
```diff
@@ -47,10 +47,12 @@ export function useProductList(query?: QueryParams) {
     };
 }
 
-export function useProductInfo(pid: number, list: ListItem[]) {
+export function useProductInfo(pid: number, list?:ListItem[]) {
     const { context } = useSession();
     const params = new URLSearchParams({ context }).toString();
-    const product = list.find(item => item.id === pid);
+
```
bc-chaz, line 53: nit: we don't need this extra line

pages/productAppExtension/[productId]/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { useRouter } from "next/router";
+import ErrorMessage from "@components/error";
+import Loading from "@components/loading";
+import { useProductInfo } from "@lib/hooks";
+
+const ProductAppExtension = () => {
+    const router = useRouter();
+    const productId = Number(router.query?.productId);
+    const list = [];
+    const { error, isLoading, product } = useProductInfo(productId, list);
```
Comment (bc-chaz) [inline pages/productAppExtension/[productId]/index.tsx:11]: since we've eliminated the previous hook, there is no longer a `list` to pass in here

```suggestion
    const { error, isLoading, product } = useProductInfo(productId);
```

pages/productAppExtension/[productId]/index.tsx
```diff
@@ -0,0 +1,39 @@
+import { H4, Panel, Text } from "@bigcommerce/big-design";
+import { useRouter } from "next/router";
+import ErrorMessage from "@components/error";
+import Loading from "@components/loading";
+import { useProductInfo } from "@lib/hooks";
+
+const ProductAppExtension = () => {
+    const router = useRouter();
+    const productId = Number(router.query?.productId);
+    const list = [];
```
Comment (bc-chaz) [inline pages/productAppExtension/[productId]/index.tsx:10]: this isn't needed since we've made `list` optional in the hook (i.e. `list?:`). Remember, adding a question mark denotes it as optional

pages/productAppExtension/[productId]/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+    const { error, isLoading, product } = useProductInfo(productId, list);
+    const { description, is_visible: isVisible, name, price, type } = product ?? {};   
+    const typeCapitalized = type?.replace(/^\w/, (c: string) => c.toUpperCase());
+    const isVisibleString = isVisible?.toString().replace(/^\w/, (c: string) => c.toUpperCase());
```
Comment (bc-chaz) [inline pages/productAppExtension/[productId]/index.tsx:14]: 🍹 this seems a bit complicated for what we're trying to do... I'd recommend either moving out the function to capitalize to a helper function or maybe just making this an easy ternary:

```suggestion
    const isVisibleString = isVisible ? 'True' : 'False';
```

pages/productAppExtension/[productId]/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+    if (isLoading) return <Loading />;
+    if (error) return <ErrorMessage error={error} />;
+  
+    return (
+        <>
+            <Panel header="Basic Information" margin="xxSmall">
```
Comment (bc-chaz) [inline pages/productAppExtension/[productId]/index.tsx:21]: I believe the issues with margin will be solved with the bug fix we looked at earlier... if we want, we can probably just override the bottom margin here, setting the first margin to `small` and then `none` for the second

```suggestion
            <Panel header="Basic Information" marginBottom="small">...</Panel>
            <Panel header="Description" margin="none">...</Panel>
```

bc-chaz (approved):

PR #160, opened 2023-05-01
chore(appex):Upgrading React and BigDesign Versions

bc-nekisham:
## What?
When trying to upgrade BigDesign theme from 0.17 to 0.18 we’re receiving a peer dependency issue requiring React to be version 18. Currently, React is version 17 in the Sample App. Will need to evaluate breaking changes moving from React 17 to React 18 and BigDesign 0.17 to BigDesign 0.18.

## Why?
Snyk upgrade https://github.com/bigcommerce/sample-app-nodejs/pull/151 

## Testing / Proof
Compiled and Loaded with React 18
Screenshot 2023-05-01 at 12 35 21 PM
Screenshot 2023-05-01 at 12 35 12 PM


package.json
```diff
@@ -20,14 +20,15 @@
...
+    "@bigcommerce/big-design-theme": "^0.18.0",
     "firebase": "^9.14.0",
     "jsonwebtoken": "^9.0.0",
     "mysql": "^2.18.1",
     "next": "^12.3.3",
     "node-bigcommerce": "github:bigcommerce/node-bigcommerce",
-    "react": "^17.0.2",
-    "react-dom": "^17.0.2",
+    "react": "18.0",
+    "react-dom": "18.0",
```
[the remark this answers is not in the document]
bc-nekisham: Updated

package.json
```diff
@@ -20,14 +20,15 @@
...
+    "@bigcommerce/big-design-icons": "^0.23.0",
+    "@bigcommerce/big-design-theme": "^0.18.0",
     "firebase": "^9.14.0",
     "jsonwebtoken": "^9.0.0",
     "mysql": "^2.18.1",
     "next": "^12.3.3",
     "node-bigcommerce": "github:bigcommerce/node-bigcommerce",
-    "react": "^17.0.2",
-    "react-dom": "^17.0.2",
+    "react": "18.0",
```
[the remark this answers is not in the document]
bc-nekisham: Updated

package.json
```diff
@@ -20,24 +20,26 @@
...
     "next": "^12.3.3",
     "node-bigcommerce": "github:bigcommerce/node-bigcommerce",
-    "react": "^17.0.2",
-    "react-dom": "^17.0.2",
-    "styled-components": "^5.3.6",
-    "swr": "^0.5.7"
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0",
+    "styled-components": "^5.3.10",
+    "swr": "^2.1.5"
```
matthewvolk, line 33: 🍹 I saw you likely updated `swr` because of the peer dependency error. Just something to keep in mind, it could be alarming for some to see a version go from `0.x.x` to `2.x.x` so suddenly, as that's two versions of potentially breaking changes that are being introduced. Just make sure to check that all the functions using `swr` are still functioning properly 👍 If anything's broken, it looks like the lowest version of SWR that supported React 18 was [1.1.0](https://github.com/vercel/swr/releases/tag/1.1.0), so you could use that version as a backup plan.

bc-nekisham: Ahhh okay, thanks! Just updated that as well.

bc-nekisham: Oh I didn't refresh the page and it stopped loading on the newer SWR version, something is going on with the way its consuming the JWT key on the newer version, the error isn't showing on the 1.1.0 though so I will leave it as that one.

matthewvolk: Awesome, great catch 👍

package.json
```diff
@@ -20,24 +20,26 @@
   },
   "dependencies": {
     "@bigcommerce/big-design": "^0.34.3",
```
bc-chaz, line 22: since we've just upgraded React, as well as the other BD packages, this might also be a good time to update the main BigDesign package to 0.36

package.json
```diff
@@ -20,24 +20,26 @@
...
-    "@bigcommerce/big-design-theme": "^0.17.0",
+    "@bigcommerce/big-design-icons": "^0.23.0",
+    "@bigcommerce/big-design-theme": "^0.18.0",
```
bc-chaz, line 24: qq: why not use `0.19` here?

package.json
bc-nekisham: Updated to 36

bc-chaz (approved):

PR #162, opened 2023-05-03
Step 1 app foundation

bc-nekisham:
## What?
When trying to upgrade BigDesign theme from 0.17 to 0.18 we’re receiving a peer dependency issue requiring React to be version 18. React was version 17 in the Sample App.

## Why?
Upgraded to React/DOM/TYPES to latest

Next to 12.3.3 

Styled Components to 5.3.10 for latest version

BigDesign to 0.36

## Testing / Proof
Tested Locally App Running 


bc-chaz:
@bc-nekisham we don't want to merge this into `main` - we should target the `step-1` branch

PR #167, opened 2023-05-18
feat(appex): APPEX-515 Create App Ext Reference

bc-nekisham:
## What?
As part of App Extensions GA, it would be nice to have a corresponding implementation to match the tutorial. This should include API calls to create the app extension (on install/ load) and delete when the app is uninstalled.  As part of this, we should build out products in a similar fashion to orders/orderId whereby the header is scrubbed for display in the side panel.

## Why?
To provide a visual example to go along with the documentation of how App Extensions work within a tutorial.

## Testing / Proof
...

https://github.com/bigcommerce/sample-app-nodejs/assets/109627308/c417f895-c3e9-4413-971f-235913dd8838

https://github.com/bigcommerce/sample-app-nodejs/assets/109627308/0d130914-c01f-45c7-b55e-e807e7df900e

### DIFF
--- a/lib/appExtensions.ts
+++ b/lib/appExtensions.ts
@@ -1,0 +1,72 @@
+//  Builds the GraphQL mutation required to create a new App Extension
+export function createAppExtension() {
+    const body = {
+        query: `
+        mutation AppExtension($input: CreateAppExtensionInput!) {
+            appExtension {
+                createAppExtension(input: $input) {
+                appExtension {
+                    id
+                    context
+                    label {
+                    defaultValue
+                    locales {
+                        value
+                        localeCode
+                        }
+                    }
+                    model
+                    url
+                    }
+                }
+            }
+        }`,
+        variables: {
+            input: {
+                context: 'PANEL',
+                model: 'PRODUCTS',
+                url: '/productAppExtension/${id}',
+                label: {
+                    defaultValue: 'View Inside Sample App',
+                    locales: [
+                        {
+                            value: 'View Inside Sample App',
+                            localeCode: 'en-US',
+                        },
+                        {
+                            value: 'Ver dentro de la aplicación de muestra',
+                            localeCode: 'es-ES',
+                        },
+                    ],
+                },
+            },
+        },
+    };
+
+    const requestBody = {
+        query: body.query,
+        variables: body.variables,
+    };
+
+    return requestBody;
+}
+
+// Builds the GraphQL query required to retrieve all App Extensions installed on a store
+export function getAppExtensions() {
+    const body = {
+        query: `
+            query {
+                store {
+                    appExtensions {
+                        edges {
+                            node {
+                                id
+                            }
+                        }
+                    }
+                }
+            }`,
+    };
+
+    return body;
+}
--- a/pages/api/auth.ts
+++ b/pages/api/auth.ts
@@ -1,4 +1,5 @@
 import { NextApiRequest, NextApiResponse } from 'next';
+import { createAppExtension } from '@lib/appExtensions';
 import { encodePayload, getBCAuth, setSession } from '../../lib/auth';
 
 export default async function auth(req: NextApiRequest, res: NextApiResponse) {
@@ -6,8 +7,34 @@ export default async function auth(req: NextApiRequest, res: NextApiResponse) {
         // Authenticate the app on install
         const session = await getBCAuth(req.query);
         const encodedContext = encodePayload(session); // Signed JWT to validate/ prevent tampering
+        const { access_token: accessToken, context } = session;
+
+        const storeHash = context?.split('/')[1] || '';
 
         await setSession(session);
+
+        /**
+         * For stores that do not have the app installed yet, create App Extensions when app is
+         * installed.
+         */
+
+        const response = await fetch(
+            `https://${process.env.API_URL}/stores/${storeHash}/graphql`,
+            {
+                method: 'POST',
+                headers: {
+                    accept: 'application/json',
+                    'content-type': 'application/json',
+                    'x-auth-token': accessToken,
+                },
+                body: JSON.stringify(createAppExtension()),
+            }
+        );
+        const { errors } = await response.json();
+        if (errors && errors.length > 0) {
+            throw new Error(errors[0]?.message);
+        }
+
         res.redirect(302, `/?context=${encodedContext}`);
     } catch (error) {
         const { message, response } = error;
--- a/pages/api/load.ts
+++ b/pages/api/load.ts
@@ -1,12 +1,16 @@
 import { NextApiRequest, NextApiResponse } from 'next';
+import { createAppExtension, getAppExtensions } from '@lib/appExtensions';
+import { getStoreToken } from '@lib/dbs/firebase';
 import { encodePayload, getBCVerify, setSession } from '../../lib/auth';
 
 const buildRedirectUrl = (url: string, encodedContext: string) => {
     const [path, query = ''] = url.split('?');
-    const queryParams = new URLSearchParams(`context=${encodedContext}&${query}`);
+    const queryParams = new URLSearchParams(
+        `context=${encodedContext}&${query}`
+    );
 
     return `${path}?${queryParams}`;
-}
+};
 
 export default async function load(req: NextApiRequest, res: NextApiResponse) {
     try {
@@ -14,7 +18,61 @@ export default async function load(req: NextApiRequest, res: NextApiResponse) {
         const session = await getBCVerify(req.query);
         const encodedContext = encodePayload(session); // Signed JWT to validate/ prevent tampering
 
+        const { sub } = session;
+        const storeHash = sub?.split('/')[1] || '';
+
+        const accessToken = await getStoreToken(storeHash);
+
         await setSession(session);
+
+        /**
+         * For stores that do not have app extensions installed yet, create app extensions when app is
+         * loaded
+         */
+
+        const existingAppExtensions = await fetch(
+            `https://${process.env.API_URL}/stores/${storeHash}/graphql`,
+            {
+                method: 'POST',
+                headers: {
+                    accept: 'application/json',
+                    'content-type': 'application/json',
+                    'x-auth-token': accessToken,
+                },
+                body: JSON.stringify(getAppExtensions()),
+            }
+        );
+
+        const { data } = await existingAppExtensions.json();
+
+        const existingAppExtensionIds = data?.store?.appExtensions?.edges;
+
+        // If there are no app extensions returned, we assume we have not
+        // installed app extensions on this store, so we must install them.
+        if (!existingAppExtensionIds?.length) {
+            const createAppExtensionRequest = await createAppExtension();
+
+            // Make fetch to create the app extensions
+            const createResponse = await fetch(
+                `${process.env.API_BASE_URL}/stores/${storeHash}/graphql`,
+                {
+                    method: 'POST',
+                    headers: {
+                        accept: 'application/json',
+                        'content-type': 'application/json',
+                        'x-auth-token': accessToken,
+                    },
+                    body: JSON.stringify(createAppExtensionRequest),
+                }
+            );
+            const response = await createResponse.json();
+
+            // Handel error if app extension creation fails
+            if (response.data === null && response.errors?.length > 0) {
+                throw new Error(response.errors[0]?.message);
+            }
+        }
+
         res.redirect(302, buildRedirectUrl(session.url, encodedContext));
     } catch (error) {
         const { message, response } = error;
--- a/package-lock.json
+++ b/package-lock.json
@@ -43,6 +43,7 @@
         "eslint-plugin-react-hooks": "^4.6.0",
         "jest": "^28.1.2",
         "jest-environment-jsdom": "^28.1.2",
+        "jest-fetch-mock": "^3.0.3",
         "typescript": "^4.7.3"
       },
       "engines": {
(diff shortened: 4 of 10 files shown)


lib/appExtensions.ts
```diff
@@ -0,0 +1,221 @@
...
+                                        id
+                                    }
+                                }
+                            }
+                        }
+                    }`,
+      }),
+    }
+  );
+  if (!response.ok) {
```
bc-zachary, line 170: 🍹 Is there a reason we're handling errors differently for `get` than for `delete`? I think the way `delete` is handled would be preferred here because it includes the error message from the server, which would help for debugging.

pages/api/load.ts
```diff
@@ -1,23 +1,61 @@
...
-import { encodePayload, getBCVerify, setSession } from '../../lib/auth';
+import { NextApiRequest, NextApiResponse } from "next";
+import { checkAppExtensionsExist } from "@lib/appExtensions";
+import { getStoreToken } from "@lib/dbs/firebase";
+import { encodePayload, getBCVerify, setSession } from "../../lib/auth";
 
 const buildRedirectUrl = (url: string, encodedContext: string) => {
-    const [path, query = ''] = url.split('?');
-    const queryParams = new URLSearchParams(`context=${encodedContext}&${query}`);
+  const [path, query = ""] = url.split("?");
```
bc-zachary, line 7: I think we prefer single quotes unless the string contains a single quote. Looks like all your quotes are double. Could be auto-linting in your IDE. This might actually be something we should add to eslint config.

bc-zachary, line 7: 🍹 Also noticing indents getting reformatted to 2 spaces. I think we're using 4 in this repo? (cc: @bc-chaz)

lib/appExtensions.ts
bc-nekisham: Updated to match error handling in get

lib/appExtensions.ts
```diff
@@ -0,0 +1,221 @@
+// Creates a new App Extension on installation of Sample App
+export async function createAppExtension(
+  storeHash: string,
+  accessToken: string
+) {
+  const response = await fetch(
+    `https://api.bigcommerce.com/stores/${storeHash}/graphql`,
```
bc-chaz, line 7: I probably wouldn't hide the api call inside of the lib folder here - I think this might actually be better as part of the API - we do something similar with checkout: https://github.com/bigcommerce/sample-app-nodejs/blob/automated-billing-beta/pages/api/checkout/%5Bpid%5D.ts#L14

lib/appExtensions.ts
```diff
@@ -0,0 +1,221 @@
...
+    defaultValue: string;
+    locales: Array<{ value: string; localeCode: string }>;
+  }>
+) {
+  // Get current App Extensions
+  const currentAppExtensions = await getAppExtensions(storeHash, accessToken);
+
+  // Get Labels of Current App Extensions
+  const currentAppExtensionLabels = await currentAppExtensions.map(
+    (edge: any) => edge.node.label
```
bc-chaz, line 133: we probably shouldn't be using any

lib/appExtensions.ts
```diff
@@ -0,0 +1,221 @@
...
+  expectedAppExtensions: Array<{
+    defaultValue: string;
+    locales: Array<{ value: string; localeCode: string }>;
+  }>
+) {
+  // Get current App Extensions
+  const currentAppExtensions = await getAppExtensions(storeHash, accessToken);
+
+  // Get Labels of Current App Extensions
+  const currentAppExtensionLabels = await currentAppExtensions.map(
```
bc-chaz, line 132: this seems odd to me that we're looking up if the app extension exists by its label rather than something like an ID... when we create the app extension initially on install, what's the `response` that we get?

lib/appExtensions.ts
```diff
@@ -0,0 +1,221 @@
...
+                        store {
+                            appExtensions {
+                                edges {
+                                    node {
+                                        id
+                                    }
+                                }
+                            }
+                        }
+                    }`,
```
bc-chaz, line 166: the spacing here looks off

lib/appExtensions.ts
```diff
@@ -0,0 +1,221 @@
...
+      `Error getting installed App Extensions: ${JSON.stringify(data.errors)}`
+    );
+  }
+  const edges = data?.data?.store?.appExtensions?.edges || [];
+
+  return edges;
+}
+
+// Deletes App Extension on Uninstall of App
+export async function deleteAppExtensions(
```
bc-chaz, line 182: I believe we're handling this already - we likely do not need to explicitly call to delete the extensions here

lib/appExtensions.ts
bc-nekisham: Removed api calls from appExtensions file

lib/appExtensions.ts
bc-nekisham: Removed delete function

lib/auth.ts
```diff
@@ -9,94 +9,94 @@ const { API_URL, AUTH_CALLBACK, CLIENT_ID, CLIENT_SECRET, JWT_KEY, LOGIN_URL } =
 // Used for internal configuration; 3rd party apps may remove
 const apiConfig: ApiConfig = {};
 if (API_URL && LOGIN_URL) {
-    apiConfig.apiUrl = API_URL;
-    apiConfig.loginUrl = LOGIN_URL;
+  apiConfig.apiUrl = API_URL;
+  apiConfig.loginUrl = LOGIN_URL;
```
bc-chaz, line 13: it looks like this file switched to 2 space indenting - I believe we're using `4` for this repo (since at the time that was more common)

pages/api/auth.ts
```diff
@@ -1,16 +1,42 @@
...
+  await setSession(session);
+
+  /**
+   * For stores that do not have the app installed yet, create App Extensions when app is
+   * installed.
+   */
+
+  const requestBody = await createAppExtension();
+
+  const response = await(fetch(`https://api.bigcommerce.com/stores/${storeHash}/graphql`, {
```
bc-chaz, line 23: we likely don't want to have the full prod api url like this for two reasons 1) it's not very reusable  and 2) perhaps more importantly, since this is a prod url, we cannot test on any of our dev/ testing environments.

I recommend reviewing what we did for checkout - https://github.com/bigcommerce/sample-app-nodejs/blob/automated-billing-beta/pages/api/checkout/%5Bpid%5D.ts#L14.

Basically, we might want to export this to a variable where we can update the base api location (e.g. `api.bigcommerce.com`) accordingly

pages/api/auth.ts
```diff
@@ -1,16 +1,42 @@
...
+  const { data, errors } = await response.json();
+  if(data === null && errors[0]?.message){
+    throw new Error(errors[0]?.message);
+  }
+
+  res.redirect(302, `/?context=${encodedContext}`);
+  } catch (error) {
+  const { message, response } = error;
+  res.status(response?.status || 500).json({ message });
+  }
```
bc-chaz, line 41: the indenting here seems weird

pages/api/auth.ts
```diff
@@ -1,16 +1,42 @@
...
+    method: "POST",
+        headers: {
+        accept: "application/json",
+        "content-type": "application/json",
+        "x-auth-token": accessToken,
+        },
+        body: JSON.stringify(requestBody),
+  }));
+  const { data, errors } = await response.json();
+  if(data === null && errors[0]?.message){
```
bc-chaz, line 33: qq: is `errors` an empty array if there are no errors? wondering if we could just check the length of errors here

pages/api/load.ts
```diff
@@ -1,20 +1,81 @@
...
+};
 
 export default async function load(req: NextApiRequest, res: NextApiResponse) {
     try {
         // Verify when app loaded (launch)
         const session = await getBCVerify(req.query);
         const encodedContext = encodePayload(session); // Signed JWT to validate/ prevent tampering
 
+        const { sub } = session;
+
```
bc-chaz, line 20: 🍹 small nit: remove extra line

pages/api/load.ts
```diff
@@ -1,20 +1,81 @@
...
+                    method: 'POST',
+                    headers: {
+                        accept: 'application/json',
+                        'content-type': 'application/json',
+                        'x-auth-token': accessToken
+                    },
+                    body: JSON.stringify(createAppExtensionRequest)
+                }
+            );
+            const { data: createAppExtensionData, errors: createAppExtensionErrors } =
```
bc-chaz, line 67: qq: why are we renaming `data` and `errors` here? it doesn't seem like these two variables have been namespaced in this section already

lib/appExtensions.ts
```diff
@@ -0,0 +1,72 @@
...
+                            appExtensions {
+                                edges {
+                                    node {
+                                        id
+                                    }
+                                }
+                            }
+                        }
+                    }`
+    };
```
bc-chaz, line 69:
```suggestion
        query: `
            query {
                store {
                    appExtensions {
                        edges {
                            node {
                                id
                            }
                        }
                    }
                }
            }`
    };
```

pages/api/auth.ts
bc-nekisham: Oh yeah it is an empty array- checking the length makes sense with that!

pages/api/load.ts
bc-nekisham: ^ changing this I renamed this when I was making changes and had another error check with a second api call that I didn't need, this is unnecessary now as its just the one instance

pages/api/auth.ts
```diff
@@ -1,13 +1,42 @@
...
+                    'content-type': 'application/json',
+                    'x-auth-token': accessToken,
+                },
+                body: JSON.stringify(requestBody),
+            }
+        );
+        const errors = await response.json();
+        if (errors.length > 0) {
+            throw new Error(errors[0]?.message);
+        }
```
matthewvolk, line 38: `await response.json()` returns a resolved promise of the body from the POST request to create an app extension. Naming the variable as `errors` is misleading, because the body could be an error response, _or a success response_. 

If your goal here is to check if the GraphQL response returned an error, I think the better way to go about this would be using the [`Response: ok` property](https://developer.mozilla.org/en-US/docs/Web/API/Response/ok)

```ts
if (!response.ok) {
  const error = await response.json();
  throw new Error(error.message); // keep in mind, error.message might be undefined if the `response.json()` object does not have a top-level message property.  
}
```

bc-nekisham: Updated and added in a default string to throw if the message is undefined

pages/api/auth.ts
```diff
@@ -1,13 +1,42 @@
...
+                    accept: 'application/json',
+                    'content-type': 'application/json',
+                    'x-auth-token': accessToken,
+                },
+                body: JSON.stringify(requestBody),
+            }
+        );
+        if (!response.ok) {
+            const error = await response.json();
+            throw new Error(error.message || 'Unknown error occurred');
```
matthewvolk, line 37: `'Unknown error occurred'` - We do know some things about the error. For example, I think a better message would describe to the app developer why their request failed. Something like that might be: <code>\`Response status code ${response.status} received from BigCommerce API when trying to create app extensions\`</code>

matthewvolk (changes requested):

pages/api/auth.ts
bc-nekisham: Ahhh okay, that makes sense and a better description!

lib/appExtensions.ts
```diff
@@ -0,0 +1,72 @@
+// Creates a new App Extension on installation of Sample App
+export async function createAppExtension() {
```
matthewvolk, line 2: Is there any reason this function needs to return a Promise? All of the code inside the function can run synchronously 🙂

lib/appExtensions.ts
```diff
@@ -0,0 +1,72 @@
...
+    const requestBody = {
+        query: body.query,
+        variables: body.variables,
+    };
+
+    return requestBody;
+}
+
+// Gets all current App Extensions installed on store
+export async function getAppExtensions() {
```
matthewvolk, line 55: Same as above - all the code in this function can run synchronously 🙂

pages/api/auth.ts
```diff
@@ -1,13 +1,45 @@
...
+        const storeHash = context?.split('/')[1] || '';
 
         await setSession(session);
+
+        /**
+         * For stores that do not have the app installed yet, create App Extensions when app is
+         * installed.
+         */
+
+        const requestBody = await createAppExtension();
```
matthewvolk, line 21: I suggest removing this line, and replacing it with the suggestion below

pages/api/auth.ts
```diff
@@ -1,13 +1,45 @@
...
+        const response = await fetch(
+            `https://${process.env.API_URL}/stores/${storeHash}/graphql`,
+            {
+                method: 'POST',
+                headers: {
+                    accept: 'application/json',
+                    'content-type': 'application/json',
+                    'x-auth-token': accessToken,
+                },
+                body: JSON.stringify(requestBody),
```
matthewvolk, line 32:
```suggestion
                body: JSON.stringify(createAppExtension());
```

lib/appExtensions.ts
```diff
@@ -0,0 +1,72 @@
+// Creates a new App Extension on installation of Sample App
```
matthewvolk, line 1:
```suggestion
// Builds the GraphQL mutation required to create a new App Extension
```

lib/appExtensions.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    const requestBody = {
+        query: body.query,
+        variables: body.variables,
+    };
+
+    return requestBody;
+}
+
+// Gets all current App Extensions installed on store
```
matthewvolk, line 54:
```suggestion
// Builds the GraphQL query required to retrieve all App Extensions installed on a store
```

lib/appExtensions.ts
bc-nekisham: Oh that was from my prior implementations so that should be removed now Thanks!

matthewvolk (approved):
Looks great 👍 There were a lot of changes we threw in at the end there, so before merging I would just do one more manual test to make sure everything still works as expected. Thanks!

bc-nekisham:
Checked change or error check change here / tested with wrong field name on query

Screenshot 2023-06-02 at 1 37 53 PM
Screenshot 2023-06-02 at 1 37 36 PM

matthewvolk (approved):
Thank you! 👍

PR #171, opened 2023-06-08
(fix): AccessToken for sql and fb db and .env url

bc-nekisham:
## What?
Fix for accessToken request in load for (fb and sql db).  Also added in API url to .env for reference

## Why?
Prior getStoreToken broke for SQL db users.

## Testing / Proof

https://github.com/bigcommerce/sample-app-nodejs/assets/109627308/6d4c054c-340b-418e-8c20-108544f4fa5e


bc-chaz (approved):
lgtm

PR #172, opened 2023-06-13
(feat): APPEX-540 check scopes when registering an app extension

MariaJose:
## What?
Add logs when app extensions scope is missing.

## Why?
Check to see if the app has the appropriate scope (app extensions) enabled to register an app extension.

## Testing / Proof
Tested locally with both mysql and firebase dbs

**Showing logs when app extensions scope is missing**

https://github.com/bigcommerce/sample-app-nodejs/assets/39739180/219e44e6-8ced-4f9b-bc11-a87ad5f0d25b

**Not showing logs when scope exists**

https://github.com/bigcommerce/sample-app-nodejs/assets/39739180/fa18f820-1171-4cbd-ae1f-4a481734951f

### DIFF
--- a/lib/appExtensions.ts
+++ b/lib/appExtensions.ts
@@ -1,5 +1,26 @@
+export const createAppExtension = async ({ accessToken, storeHash }: { accessToken: string, storeHash: string }) => {
+    const response = await fetch(
+        `https://${process.env.API_URL}/stores/${storeHash}/graphql`,
+        {
+          method: "POST",
+          headers: {
+            accept: "application/json",
+            "content-type": "application/json",
+            "x-auth-token": accessToken,
+          },
+          body: JSON.stringify(createAppExtensionMutation()),
+        }
+    );
+
+    const { errors } = await response.json();
+  
+    if (errors && errors.length > 0) {
+      throw new Error(errors[0]?.message);
+    }
+}
+
 //  Builds the GraphQL mutation required to create a new App Extension
-export function createAppExtension() {
+export function createAppExtensionMutation() {
     const body = {
         query: `
         mutation AppExtension($input: CreateAppExtensionInput!) {
--- a/lib/dbs/mysql.ts
+++ b/lib/dbs/mysql.ts
@@ -91,3 +91,10 @@ export async function getStoreToken(storeHash: string) {
 export async function deleteStore({ store_hash: storeHash }: SessionProps) {
     await query('DELETE FROM stores WHERE storeHash = ?', storeHash);
 }
+
+export async function hasAppExtensionsScope(storeHash: string): Promise<boolean> {
+    const scopes = await query('SELECT scope FROM stores WHERE storeHash = ?', storeHash);
+    const row = scopes[0];
+
+    return row?.scope?.includes('store_app_extensions_manage') ?? false;
+}
--- a/pages/api/auth.ts
+++ b/pages/api/auth.ts
@@ -3,6 +3,8 @@ import { createAppExtension } from '@lib/appExtensions';
 import { encodePayload, getBCAuth, setSession } from '../../lib/auth';
 
 export default async function auth(req: NextApiRequest, res: NextApiResponse) {
+    const isAppExtensionsScopeEnabled = req.query.scope.includes('store_app_extensions_manage');
+
     try {
         // Authenticate the app on install
         const session = await getBCAuth(req.query);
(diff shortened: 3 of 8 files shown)


lib/appExtensions.ts
bc-nekisham, line 1: There was a discussion on having the API calls outside of the lib folder/appExt file when I added in the prior GraphQL mutation here https://github.com/bigcommerce/sample-app-nodejs/pull/167#discussion_r1202754533 Chaz would need to way in on that but besides that it looks good to merge!

types/db.ts
```diff
@@ -19,4 +19,5 @@ export interface Db {
     getStoreToken(storeId: string): Promise<string> | null;
     deleteStore(session: SessionProps): Promise<void>;
     deleteUser(session: SessionProps): Promise<void>;
+    hasAppExtensionsScope(storeHash: string): Promise<boolean> | boolean;
```
rtalvarez, line 22: I wonder if we should consider making this only `Promise<boolean>` and not `Promise<boolean> | boolean` to keep the interface consistent, so instead of returning `false` above on `mysql.ts` we should probably return a rejected promise instead?

rtalvarez (commented):
Side note, not really for your PR but, should we consider doing some sort of automatic linting on this repo, and probably use our bigcommerce lint rules? I noticed some of the lines are missing `;`, for example

MariaJose:
> Side note, not really for your PR but, should we consider doing some sort of automatic linting on this repo, and probably use our bigcommerce lint rules? I noticed some of the lines are missing `;`, for example

Agree, we should be using bc lint rules 👍🏼  I'll create a ticket for that

pages/api/auth.ts
```diff
@@ -18,21 +20,10 @@ export default async function auth(req: NextApiRequest, res: NextApiResponse) {
          * installed.
          */
 
-        const response = await fetch(
```
bc-zachary, line 21: More dry 👍 . I think we should still keep the fetch in a file within the `pages/api/` directory for the sake of organization. Maybe we could move the `createAppExtension` function out of `lib/appExtensions.ts` and into it's own file `pages/api/auth/createAppExtension.ts` and rename this file to `pages/api/auth/index.ts`

MariaJose: It's a great idea 👍🏼 , but if we add it in the `api/auth` folder next.js is going to considere it as an api route, and in case someone access is going to crash. 

Should it be good to place `createAppExtension`  inside `/lib/auth` file ? Noticed  `getAuth` lives there, and we are importing it into our api route --> `api/auth.ts` file 

Otherwise, we could continue keeping individualy the fetch in the corresponding api pages and not using dry, up to us 👍🏼

lib/appExtensions.ts
MariaJose: @bc-chaz since I want to prevent duplication when creating app extension I created `createAppExtension` , wondering if we could place it in `lib/auth.ts` instead? Or should we continue using the API calls in corresponding api pages?

pages/api/auth.ts
bc-zachary: Ah I see what you're saying. My thinking was about keeping all external requests within the api directory, but you're right that we are making requests via node-bigcommerce from lib. I think this is okay.

lib/dbs/mysql.ts
```diff
@@ -91,3 +91,14 @@ export async function getStoreToken(storeHash: string) {
...
     await query('DELETE FROM stores WHERE storeHash = ?', storeHash);
 }
+
+export async function hasAppExtensionsScope(storeHash: string): Promise<boolean> {
+    const scopes = await query('SELECT scope FROM stores WHERE storeHash = ?', storeHash);
+    const row = scopes[0];
+
+    if (!row) {
+       return false;
+    }
```
bc-zachary, line 101: 🍹 Can probably one-line this `if` statement.
```suggestion
    if (!row) return false;
```

bc-zachary (approved):

bc-chaz (approved):

lib/dbs/mysql.ts
```diff
@@ -91,3 +91,10 @@ export async function getStoreToken(storeHash: string) {
 export async function deleteStore({ store_hash: storeHash }: SessionProps) {
     await query('DELETE FROM stores WHERE storeHash = ?', storeHash);
 }
+
+export async function hasAppExtensionsScope(storeHash: string): Promise<boolean> {
+    const scopes = await query('SELECT scope FROM stores WHERE storeHash = ?', storeHash);
+    const row = scopes[0];
+
+    return row?.scope?.includes('store_app_extensions_manage') ?? false;
```
MariaJose, line 99: @bc-chaz added it in one line 👍🏼

PR #198, opened 2023-08-14
fix readme syntax that caused suboptimal markdown rendering

slsriehl:
## What?
* fix readme syntax that caused suboptimal markdown rendering

## Why?
* suboptimal markdown parser did not properly pick up the end of a fenced code block, which made part of the readme appear as though instructions were code blocks

## Testing / Proof
~~https://github.com/slsriehl/sample-app-nodejs/blob/082079baaecac920cf599367db32f27201c65fa4/README.md~~
https://github.com/slsriehl/sample-app-nodejs/blob/b4a766afe447105420e809d7ff8fcfba6b94061e/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ This starter app includes all the files necessary to get started with a basic, h
 To get the app running locally, follow these instructions:
 
 1. [Use Node 16+ and NPM 8+](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm#checking-your-version-of-npm-and-node-js). To check the running versions, use the following commands:
-   
+
 ```shell
 node -v
 npm -v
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,19 +23,21 @@ npm install
...
-    You can use `npm` to install ngrok:
+   You can use `npm` to install ngrok:
 
 ```shell
 npm install -g ngrok
 ```
 
-    Alternatively, MacOS users can use the [homebrew](https://brew.sh/) package manager:
+&#13;
+   Alternatively, MacOS users can use the [homebrew](https://brew.sh/) package manager:
```
rtalvarez, line 33:
```suggestion
Alternatively, MacOS users can use the [homebrew](https://brew.sh/) package manager:
```

rtalvarez: If we remove these spaces we can probably delete the manual `&#13;` , as far as I can see in my IDE the rendered result is the same

rtalvarez (approved):

PR #205, opened 2023-09-06
Appex 628

bc-chaz:
## What?
Updates documentation to support running production builds. Also converts database URL from clearDB to a generic URL. This PR also addresses an issue with mysql import. 

## Why?
developer/ fly.io feedback

## Testing / Proof
locally, confirmed build passed and app functions

### DIFF
--- a/.env-sample
+++ b/.env-sample
@@ -24,6 +24,9 @@ FIRE_PROJECT_ID={firebase project id}
 
 # If using mysql, enter your config here
 
+# DATABASE_URL={mysql://db_address}
+# DB config can also be entered manually:
+
 MYSQL_HOST={mysql host}
 MYSQL_DATABASE={mysql database name}
 MYSQL_USERNAME={mysql username}
--- a/.env-sample
+++ b/.env-sample
@@ -22,10 +22,9 @@ FIRE_API_KEY={firebase key}
 FIRE_DOMAIN={firebase domain}
 FIRE_PROJECT_ID={firebase project id}
 
-# If using mysql, enter your config here
+# If using mysql, You can use a database URL or enter multiple configuration variables. Comment out the variables you don't use by adding the `#` character to the beginning of the line.
 
 # DATABASE_URL={mysql://db_address}
-# DB config can also be entered manually:
 
 MYSQL_HOST={mysql host}
 MYSQL_DATABASE={mysql database name}
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 This starter app includes all the files necessary to get started with a basic, hello world app. This app uses NextJS, BigDesign, Typescript, and React.
 
-## Running the app
+## Running the app in development
 
 To get the app running locally, follow these instructions:
 
@@ -81,13 +81,15 @@ npm run dev
 
 > If `ngrok` expires, update the callbacks in steps 4 and 7 with the new `ngrok_url`. You can learn more about [persisting ngrok tunnels longer (ngrok)](https://ngrok.com/docs/getting-started/#step-3-connect-your-agent-to-your-ngrok-account).
 
-11. To create and run a production build, an optimized version of the code must be built and then run:
+11. Consult our developer documentation to [install and launch the app](https://developer.bigcommerce.com/api-docs/apps/quick-start#install-the-app).
+
+## Production Builds
+
+In production, you must build and run optimized version of the code. Use the following commands to get started:
+
+> When you run the `start` script, specify a port number using the `-p` flag. Otherwise, the process will fail.
 
 ```shell
 npm run build
 npm run start -p 3000
 ```
-
-   In order to start in production mode, a port number must be specified by using the `-p` flag.
-
-12. Consult our developer documentation to [install and launch the app](https://developer.bigcommerce.com/api-docs/apps/quick-start#install-the-app).
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ npm run dev
 
 11. Consult our developer documentation to [install and launch the app](https://developer.bigcommerce.com/api-docs/apps/quick-start#install-the-app).
 
-## Production Builds
+## Production builds
 
 In production, you must build and run optimized version of the code. Use the following commands to get started:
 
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -81,4 +81,13 @@ npm run dev
 
 > If `ngrok` expires, update the callbacks in steps 4 and 7 with the new `ngrok_url`. You can learn more about [persisting ngrok tunnels longer (ngrok)](https://ngrok.com/docs/getting-started/#step-3-connect-your-agent-to-your-ngrok-account).
 
-11. Consult our developer documentation to [install and launch the app](https://developer.bigcommerce.com/api-docs/apps/quick-start#install-the-app).
+11. To create and run a production build, an optimized version of the code must be built and then run:
```
bc-chaz, line 84: @slsriehl would you mind taking a pass at this 🙏

rtalvarez: I'm a bit confused about why we're suddenly talking about production, since the readme up until this point is mostly focused on running locally 

Also, I think this is a bit awkward, because if you're following the guide, you'd run `ngrok http 3000` which uses up that port, so running `npm run start -p 3000` would clearly collide. 

We should either change that to point to somewhere else, or just instruct people to run `npm run dev` instead of start. Dev will fallback to a different port if the default is in use, whereas `start` will crash with `EADDRINUSE`

.env-sample
```diff
@@ -24,6 +24,9 @@ FIRE_PROJECT_ID={firebase project id}
 
 # If using mysql, enter your config here
```
slsriehl, line 25:
```suggestion
# If using mysql, enter your config here. You can use a database URL or enter multiple configuration variables.
```

.env-sample
```diff
@@ -24,6 +24,9 @@ FIRE_PROJECT_ID={firebase project id}
...
 
+# DATABASE_URL={mysql://db_address}
+# DB config can also be entered manually:
```
slsriehl, line 28:
```suggestion
```

.env-sample
```diff
@@ -24,6 +24,9 @@ FIRE_PROJECT_ID={firebase project id}
 
 # If using mysql, enter your config here
 
+# DATABASE_URL={mysql://db_address}
```
slsriehl, line 27:
```suggestion
DATABASE_URL={mysql://db_address}
```

README.md
slsriehl: @rtalvarez I believe `ngrok http 3000` proxies whatever process is running on port 3000; if I recall correctly, ngrok runs on one of the 4000 ports.

README.md
```diff
@@ -81,4 +81,13 @@ npm run dev
...
+
+```shell
+npm run build
+npm run start -p 3000
```
slsriehl, line 88: If I recall correctly, `start` works without `run`.

```suggestion
npm start -p 3000
```

README.md
slsriehl: @bc-chaz I agree with @rtalvarez that production info should be presented separately.

slsriehl, line 84:
```suggestion
11. Consult our developer documentation to [install and launch the app](https://developer.bigcommerce.com/api-docs/apps/quick-start#install-the-app).
```

README.md
```diff
@@ -81,4 +81,13 @@ npm run dev
 
 > If `ngrok` expires, update the callbacks in steps 4 and 7 with the new `ngrok_url`. You can learn more about [persisting ngrok tunnels longer (ngrok)](https://ngrok.com/docs/getting-started/#step-3-connect-your-agent-to-your-ngrok-account).
 
-11. Consult our developer documentation to [install and launch the app](https://developer.bigcommerce.com/api-docs/apps/quick-start#install-the-app).
+11. To create and run a production build, an optimized version of the code must be built and then run:
+
```
slsriehl, line 85:
```suggestion
## Production builds

In production, you must build and run optimized version of the code. Use the following commands to get started:

> When you run the `start` script, specify a port number using the `-p` flag. Otherwise, the process will fail.

```

README.md
```diff
@@ -81,4 +81,13 @@ npm run dev
...
+```shell
+npm run build
+npm run start -p 3000
+```
+
+   In order to start in production mode, a port number must be specified by using the `-p` flag.
```
slsriehl, line 91: delete the rest of this stuff.

My changes depend on changing the first H2 in the doc to the following:

```
## Running the app in development
```

README.md
rtalvarez: > @rtalvarez I believe `ngrok http 3000` proxies whatever process is running on port 3000; if I recall correctly, ngrok runs on one of the 4000 ports.

Ahh of course! My bad

bc-chaz: @rtalvarez @slsriehl the main impetus behind this is from fly.io who mentioned that this lack of information was problematic when trying to start and containerize the sample app. BC is looking at fly.io as a replacement for Heroku so we would likely want to include this information

bc-chaz: pls see comments above

README.md
bc-chaz: it can be run either way - for consistency with all of the other commands, and to mitigate mistakes elsewhere, I included it

.env-sample
bc-chaz: we likely want to keep the variables as the default - the nature of this PR should not change what we've been doing

bc-nekisham (approved):
lgtm

README.md
```diff
@@ -82,3 +82,14 @@ npm run dev
 > If `ngrok` expires, update the callbacks in steps 4 and 7 with the new `ngrok_url`. You can learn more about [persisting ngrok tunnels longer (ngrok)](https://ngrok.com/docs/getting-started/#step-3-connect-your-agent-to-your-ngrok-account).
 
 11. Consult our developer documentation to [install and launch the app](https://developer.bigcommerce.com/api-docs/apps/quick-start#install-the-app).
+
+## Production Builds
```
slsriehl, line 86:
```suggestion
## Production builds
```

slsriehl (approved):
I made one small suggestion for consistency. otherwise, LGTM.

rtalvarez (approved):

rtalvarez (approved):

